PR #2, opened 2018-04-15
Change the default unit system to MKS

jzuhone:
This PR changes the default unit system of `unyt` to SI/MKS. This requires touching nearly all the physical ratios, most of the physical constants, and the unit lookup table, as well as a bunch of other code.

It also:

* Adds candela (dimensions of luminous intensity). 
* Adds some derived units related to candela.
* Adds units for magnetic flux in MKS and CGS.

I think it's probably appropriate to also add to the docs a bit of a discussion about the SI system, haven't done that yet. Should probably also add some more testing coverage which uses these new units and dimensions.


unyt/unit_systems.py
```diff
@@ -178,13 +199,16 @@ def __repr__(self):
...
-cgs_ampere_unit_system = UnitSystem('cgs-ampere', 'cm', 'g', 's',
-                                    current_mks_unit='A')
-cgs_ampere_unit_system["energy"] = "erg"
-cgs_ampere_unit_system["specific_energy"] = "erg/g"
-cgs_ampere_unit_system["pressure"] = "dyne/cm**2"
-cgs_ampere_unit_system["force"] = "dyne"
-cgs_ampere_unit_system["magnetic_field_cgs"] = "gauss"
-cgs_ampere_unit_system["charge_cgs"] = "esu"
-cgs_ampere_unit_system["current_cgs"] = "statA"
+#: The internal base unit system: CGS with extra SI units
```
ngoldbaum: Possibly crazy idea: why don't we just use SI?

That'll make integrating with unyt back inside yt slightly harder but that's probably worth the effort to make `unyt` easier to use for a random non-astronomer user. You'd be able to work with CGS units pretty easily anyway.

Not something that needs to happen in this PR of course.

unyt/dimensions.py
```diff
@@ -27,12 +27,16 @@
 angle = Symbol("(angle)", positive=True)
 #: current_mks
 current_mks = Symbol("(current_mks)", positive=True)
+#: amount
+amount = Symbol("(amount)", positive=True)
```
ngoldbaum: I'm not sure how I feel about needing to define this. There are certainly situations where it's convenient to think of a mol as just a number. I don't do chemistry calculations often enough to have a strong opinion though.

unyt/unit_symbols.py
```diff
@@ -179,6 +179,12 @@
 Pa = pascal = quan(1.0, "Pa")
 #: sievert
 Sv = sievert = quan(1.0, "Sv")
+#: candela
+cd = Candela = candela = quan(1.0, "cd")
+#: lumen
+lm = Lumen = lumen = quan(1.0, "lm")
+#: lux
+lx = Lux = lux = quan(1.0, 'lx')
```
ngoldbaum: Add [Lambert](https://en.wikipedia.org/wiki/Lambert_(unit)) here. Maybe some of the other luminance units as well?

unyt/unit_object.py
```diff
@@ -625,8 +626,11 @@ def get_base_equivalent(self, unit_system="cgs"):
         unit_system = unit_system_registry[str(unit_system)]
         try:
             units_string = unit_system[self.dimensions]
-        except MissingMKSCurrent:
-            units_string = em_conversions[self.dimensions][0]
+        except MissingMKSUnit:
+            if self.dimensions in em_conversions:
+                units_string = em_conversions[self.dimensions][0]
+            else:
+                raise UnitsNotReducible(self, unit_system.name)
```
ngoldbaum: If you add luminance and amount to the other unit systems, you don't need to raise this error, you could just convert the data.

Here's my suggestion, for CGS use Lambert as the default luminance unit. Otherwise use candelas (unless there's an obvious other choice for the other unit systems?). Arguing for this: we also use "radian" as the default angle unit for all the other unit systems. As far as I can tell the units for luminance [are weird](https://twitter.com/njgoldbaum/status/984156999792209920).

Is it really helpful to raise this error? If we make this one concession to keeping things more SI-like by default I think it will be easier to use the library.

ngoldbaum:
Also sorry it looks like the tests are broken. That might be my fault. I pushed a bunch of changes to the equivalency system last night that I guess you needed to rebase on?

On the plus side there are in-place equivalencies now :) And lots more tests. There are a couple bugfixes that I want to backport for yt.units that the new tests caught.

unyt/unit_systems.py
jzuhone: To be honest, I think that this is what we should do. I suspect that integrating back into yt will actually not be that hard--it may actually be all internal to unyt. We'd have to change the default base units and the `unit_lookup_table`, and maybe one or two other things.

unyt/equivalencies.py
```diff
@@ -466,29 +473,52 @@ def _convert(self, x, new_dims):
...
+    if unit_str[:2] == 'da':
+        possible_prefix = 'da'
+        prefix_len += 1
+    if possible_prefix in unit_prefixes:
+        base_unit = unit_str[prefix_len:]
+    else:
+        base_unit = unit_str
+    return base_unit
+
+def check_em_conversion(units):
```
ngoldbaum: this should be prefixed with an underscore so it doesn't end up in the public API

ngoldbaum: Does this function have much overhead? It looks like it gets called during all of the unit conversion functions so optimally it would be very cheap and not add much overhead to those function.

unyt/array.py
```diff
@@ -893,14 +905,16 @@ def in_base(self, unit_system="cgs", equivalence=None, **kwargs):
         >>> print(E.in_base("mks"))
         2.5e-07 W
         """
-        try:
+        em_units, em_us = check_em_conversion(self.units)
+        if em_units is not None and em_us == unit_system:
+            to_units = em_units
+            equivalence = em_us
```
ngoldbaum: clever :)

unyt/tests/test_unyt_array.py
```diff
@@ -1544,15 +1553,13 @@ def test_modified_unit_division():
...
+    reg1.modify('g', 50)
 
-    a = unyt_quantity(3, 'm', registry=reg1)
-    b = unyt_quantity(3, 'm', registry=reg2)
+    a = unyt_quantity(3, 'g', registry=reg1)
+    b = unyt_quantity(3, 'g', registry=reg2)
 
     ret = a/b
-    assert ret == 0.5
+    assert ret == 50000.
```
ngoldbaum: I think I'm being dumb, but is this correct?

ngoldbaum: Ah, it's because now the `base_value` of gram is .001, as it should be. Apologies for the dumbness.

I think this is ready to merge? Any objections?

jzuhone: I don't have any, but I wrote the PR, so...

We should maybe ping @MatthewTurk @chummels @brittonsmith @Xarthisius

jzuhone: @MatthewTurk @chummels @brittonsmith @Xarthisius sorry to ping you again, but I just updated the title and description of the PR, so if you started looking it in the past few minutes, you should read the title and description again--the changes are pretty important.

Xarthisius: Asking a bunch of astronomers to abandon cgs?! That's bold! :D 
Sounds good to me!

jzuhone: When we decide to make `yt` depend on `unyt` in yt 4.0, we should be able to maintain cgs as the default in `yt` with only a little bit of trouble.

ngoldbaum: Yup, another thing we might eventually do is make the default unit system used by e.g. `in_base` and friends is configurable.

ngoldbaum: > yt will potentially use unyt as a dependency and it will effectively make yt's units to default to MKS). Is this right?

I'd replace potentially with almost definitely :)

> since most of the user-base are astronomers who will want to continue having cgs as their default unit system.

Yup. I'm certain the astrophysics simulation domain context will default to CGS units.

ngoldbaum:
OK, I'm merging this.

@jzuhone thank you so much for your effort on this, it will make `unyt` a higher quality library :)

PR #5, opened 2018-05-11
Adding more imperial units and fixing "inch"

jzuhone:
So I defined the "inch" in the `default_unit_symbol_lut` as "in", which is bad for obvious reasons. This fixes that. It also adds other imperial units. 

I also changed `MeV` and `GeV` in `unit_symbols` to be `Unit` because it isn't obvious why they should be different.


unyt/_unit_lookup_table.py
```diff
@@ -108,6 +111,14 @@
...
     "atm": (pascal_per_atm, dimensions.pressure, 0.0, r"\rm{atm}"),
     "hp": (watt_per_horsepower, dimensions.power, 0.0, r"\rm{hp}"),
+    "oz": (kg_per_pound/16.0, dimensions.mass, 0.0, r"\rm{oz}"),
+    "ton": (kg_per_pound*2000.0, dimensions.mass, 0.0, r"\rm{ton}"),
+    "slug": (kg_per_pound*standard_gravity_m_per_s2/m_per_ft,
+             dimensions.mass, 0.0, r"\rm{slug}"),
+    "cal": (4.184, dimensions.energy, 0.0, r"\rm{cal}"),
+    "BTU": (1055.0559, dimensions.energy, 0.0, r"\rm{BTU}"),
+    "psi": (kg_per_pound*standard_gravity_m_per_s2/m_per_inch**2,
+            0.0, r"\rm{psi}"),
```
ngoldbaum: I think you forgot the dimension on this one

unyt/_unit_lookup_table.py
```diff
@@ -95,10 +97,11 @@
...
 
     # Imperial and other non-metric units
-    "in": (0.0254, dimensions.length, 0.0, r"\rm{in}"),
-    "ft": (0.3048, dimensions.length, 0.0, r"\rm{ft}"),
-    "yd": (0.9144, dimensions.length, 0.0, r"\rm{yd}"),
-    "mile": (1609.344, dimensions.length, 0.0, r"\rm{mile}"),
+    "inch": (m_per_inch, dimensions.length, 0.0, r"\rm{in}"),
+    "ft": (m_per_ft, dimensions.length, 0.0, r"\rm{ft}"),
+    "yd": (m_per_ft*3.0, dimensions.length, 0.0, r"\rm{yd}"),
+    "mile": (m_per_ft*5280.0, dimensions.length, 0.0, r"\rm{mile}"),
```
ngoldbaum: Can you leave this as `1609.344`? Otherwise it introduces round-off error level noise in common conversions like mile to km. `1,609.344` is the exact conversion.

ngoldbaum: Similarly for `ft` and `yd` actually, the constants there were hard-coded in there are all exact conversions.

PR #6, opened 2018-05-11
almost 100% test coverage, remove MKS/CGS equivalences in favor of direct conversions

ngoldbaum:
I realize this is a huge pull request. I can definitely break this up if that would help.

This gets us to 3 lines away from 100% test coverage. The remaining untested lines are either too difficult to test (e.g. I'd need to implement a dummy numpy ufunc) or are in code i've copy/pasted from sympy and I'd prefer to leave unchanged for simplicity.

There's a number of places where I've deleted code. It may look like those places were adding functionality but according to the test coverage those lines were never actually executed. In all cases I made a good-faith effort to try to trigger the deleted lines and usually I found that logic elsewhere made it impossible to reach that line.

I removed `UnitSystemConstants` since as far as I could see it wasn't actually used anywhere. If there was an intended purpose for that I'm happy to add it back along with a test and some docs.

The `Unit` class is no longer a subclass of sympy's `Expr` class. This makes the code simpler to understand by making it clearer where the sympy "magic" happen (in the `expr` attribute of the `Unit` class, not in the `Unit` class itself).


unyt/array.py
```diff
@@ -232,11 +231,7 @@ def _coerce_iterable_units(input_object):
                 raise IterableUnitCoercionError(input_object)
             # This will create a copy of the data in the iterable.
             return unyt_array(input_object)
-        return np.asarray(input_object)
-    else:
-        if isinstance(input_object, Unit):
-            return unyt_quantity(1.0, input_object)
-        return np.asarray(input_object)
+    return np.asarray(input_object)
```
jzuhone: I assume this "just works" now without the need for the logic under `else`?

unyt/array.py
```diff
@@ -246,14 +241,6 @@ def _unit_repr_check_same(my_units, other_units):
     is compatible with this quantity. Returns Unit object.
 
     """
-    equiv_dims = em_dimensions.get(my_units.dimensions, None)
```
jzuhone: I assume we never get here anymore, correct?

unyt/array.py
```diff
@@ -516,15 +504,6 @@ def __new__(cls, input_array, input_units=None, registry=None, dtype=None,
             if registry is not None:
                 obj.units.registry = registry
             return obj
-        if input_array is NotImplemented:
-            return input_array.view(cls)
-        if registry is None and isinstance(input_units, (str, bytes)):
```
jzuhone: Will we have a check for this once we start using `unyt` in `yt`?

unyt/array.py
```diff
@@ -1178,10 +1151,6 @@ def to_astropy(self, **kwargs):
         >>> data.to_astropy()
         <Quantity [3., 4., 5.] g / cm3>
         """
-        if _astropy.units is None:
```
jzuhone: What happens if AstroPy isn't installed?

unyt/array.py
ngoldbaum: Yup, this case is caught by `Unit.__mul__` and `Unit.__rmul__`. The `Unit` class has a higher `__array_priority__` than `unyt_array` so `Unit`'s `__mul__` and `__rmul__` wins over `unyt_array`.

unyt/array.py
```diff
@@ -1661,15 +1624,8 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
                         raise InvalidUnitOperation(
                             "Quantities with units of Fahrenheit or Celsius "
                             "cannot by multiplied, divided, subtracted or "
-                            "added.")
+                            "added with data that has different units.")
                     inp1 = np.asarray(inp1)*conv
-                else:
-                    if ((u0.base_offset and u0.dimensions is temperature or
```
jzuhone: Are we allowing this now?

unyt/unit_object.py
```diff
@@ -722,6 +696,84 @@ def latex_representation(self):
...
+    "gauss": ("T", 1.0e-4, "mks"),
+    "G": ("T", 1.0e-4, "mks"),
+    "Mx": ("Wb", 1.0e-8, "mks"),
+    "statA": ("A", 10.0/speed_of_light_cm_per_s, "mks"),
+    "statV": ("V", 1.0e8/speed_of_light_cm_per_s, "mks"),
+    "statohm": ("ohm", speed_of_light_cm_per_s**2*1.0e-9, "mks"),
+}
+
+
+def _em_conversion(orig_units, conv_data, to_units=None, unit_system=None):
```
jzuhone: It looks like you're trying to do the more complicated EM conversions here still. Is that wrong?

unyt/array.py
ngoldbaum: yup, this section is now handled by `_check_em_conversion`

unyt/array.py
ngoldbaum: Yeah, I removed this because the error message is highly yt-specific. We could try to massage the error message, but then that's less useful for yt users.

Another way to do it would be to add the error in a wrapper around `unyt_array` in yt. Although that would mean we'd need to keep using that wrapper going forward instead of just exposing `unyt_array`.

I'm open to suggestions about how to do this. I'd prefer not to leave the error in as-is though because it only makes sense in the context of yt.

jzuhone: Happy to just take the error out now--can you file an issue to remind us to fix this for yt-4.0?

unyt/array.py
ngoldbaum: The error message below actually never gets printed because doing `_astropy.units` raises an `ImportError` from the on-demand import machinery. I could have made this check do `if isinstance(_astropy, NotAModule)` instead, but decided against that because the error message from the on-demand import machinery is perfectly good.

unyt/array.py
ngoldbaum: https://github.com/yt-project/yt/issues/1785

unyt/unit_object.py
ngoldbaum: I guess it's not enforced in the function, but `to_units` will only be `None` if the `unit_system` is also not `None`. This happens in the call to `_em_conversion` in `in_base` and `get_base_equivalent`.

jzuhone: Can we get a docstring or some comments here to explain a bit the steps of this function? I know this is internal, but for future reference it would help.

ngoldbaum: Done. Let me know if you have comments about the docstrings I added.

unyt/unit_object.py
```diff
@@ -722,6 +696,111 @@ def latex_representation(self):
...
+    to_units will be None an we determine the units to conver to based on
+    the unit system we want to convert the data to.
+
+    This function does not do any error checking as that should befully handled
+    by _check_em_conversion.
+    """
+    new_expr = orig_units.copy().expr
+    inter_expr = orig_units.copy().expr
+    base_value = orig_units.base_value
+    for orig_unit in conv_data:
```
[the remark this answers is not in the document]
ngoldbaum: OK, simplified things a bit. Is this clearer now?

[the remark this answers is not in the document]
jzuhone: Yes--sorry for being a PITA.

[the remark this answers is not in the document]
ngoldbaum: That's totally fair, thanks for the review!

ngoldbaum:
@jzuhone :shipit: merge ok?

jzuhone:
Yeah, I just hadn't checked for the tests passing yet

ngoldbaum:
OK, now the remaining TODOs are to rerun the benchmarks and finish the docs. If there are any regressions in the benchmarks I'll send one final performance PR. Otherwise the paper will get submitted as soon as the docs are done.

Thanks so much for your help getting the library out the door John.

PR #8, opened 2018-05-22
Finish the docs

ngoldbaum:
Leaving this open in case anyone wants to take a look at it.

Going to start drafting the paper now. The paper will include the performance comparison.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -763,14 +763,162 @@ As for arrays with different units, for operation between two arrays with units 
 Writing Data with Units to Disk
 -------------------------------
 
-Pickles
-*******
+The :mod:`unyt` library has support for serializing data stored in a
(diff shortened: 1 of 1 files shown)


docs/usage.rst
```diff
@@ -763,14 +763,167 @@ As for arrays with different units, for operation between two arrays with units
...
+in a physical unit system, then you only need to save the unit name as a string
+and treat the array data you are trying to save as a regular numpy array, as in
+this example:
+
+  >>> import numpy as np
+  >>> import os
+  >>> from unyt import cm
+  ...
+  >>> data = [1, 2, 3]*cm
+  >>> np.save('my_data_cm.npy', data)
```
jzuhone: Not necessary to do now, but I'm wondering if we ought to have a function that does this for you.

ngoldbaum: Probably worth adding. I’m not sure offhand how extensible the .npy format is.

PR #10, opened 2018-05-29
Make lack of dependencies a bit clearer

MatthewTurk:
I thought it might be helpful to note that there aren't any dependencies that are too heavy -- especially yt itself.

### DIFF
--- a/README.rst
+++ b/README.rst
@@ -23,6 +23,9 @@ unyt
 
 A package for handling numpy arrays with units
 
+This package only depends on `numpy`, `sympy`, `six` and a backport of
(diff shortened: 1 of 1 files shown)


README.rst
```diff
@@ -23,6 +23,9 @@ unyt
 
 A package for handling numpy arrays with units
 
+This package only depends on `numpy`, `sympy`, `six` and a backport of
+`lru_cache`.  Notably, it does *not* depend on `yt`.
```
ngoldbaum: Can you say "and, on Python 2.7, `backports.functools_lru_cache`"

PR #22, opened 2018-06-28
Use dimensions of units to do CGS <-> MKS conversions.

ngoldbaum:
This PR refactors the CGS <-> MKS conversions to be based around dimensions. This makes it possible to convert code magnetic units between unit systems (or any unit that has the same dimensions as gauss or tesla).

This fixes issues with converting code E&M magnetic units between unit systems and I've added a test to capture this use case.

This came up in the context of getting yt to use unyt under the hood.

### DIFF
--- a/unyt/unit_object.py
+++ b/unyt/unit_object.py
@@ -86,6 +86,15 @@ global_dict = {
 }
 
 
+def _sanitize_unit_system(unit_system, obj):
+    from unyt.unit_systems import unit_system_registry
+    if hasattr(unit_system, "unit_registry"):
+        unit_system = unit_system.unit_registry.unit_system_id
+    elif unit_system == "code":
+        unit_system = obj.registry.unit_system_id
+    return unit_system_registry[str(unit_system)]
+
+
 def _auto_positive_symbol(tokens, local_dict, global_dict):
     """
     Inserts calls to ``Symbol`` for undefined variables.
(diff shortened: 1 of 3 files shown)


unyt/unit_object.py
```diff
@@ -711,22 +718,26 @@ def latex_representation(self):
...
+    dims.electric_potential_mks: (
+        dims.electric_potential_cgs, "statV",
+        1.0e-8*speed_of_light_cm_per_s),
+    dims.electric_potential_cgs: (
+        dims.electric_potential_mks, "V", 1.0e8/speed_of_light_cm_per_s),
+    dims.resistance_mks: (
+        dims.resistance_cgs, "statohm", 1.0e9/(speed_of_light_cm_per_s**2)),
+    dims.resistance_cgs: (
+        dims.resistance_mks, "ohm", 1.0e-9*speed_of_light_cm_per_s**2)
 }
```
jzuhone: I actually tried this before and it didn't work--and it was because two of these dimensions actually reduced to the same symbol, so one of them got overwritten in the dictionary. But now I can't remember which two....

ngoldbaum: I think it was Maxwell. I decided that magnetic field strength was more important so I removed support for automatic conversions of magnetic flux.

jzuhone: Ah, fair enough.

PR #31, opened 2018-07-14
make UnitRegistry.__contains__ a bit smarter 

### LINKED ISSUE #30: Ability to check if string is a valid unit (in any unit system)
* unyt version: 1.0.4
* Python version: 2.7-3.6
* Operating System: MacOSX

### Description

It would be awesome to be able to check if a string is a valid unit by something like:

```
if 'something' in ureg:
    ...
```

### What I Did

Currently, you can get this by using try, except, but it would be useful out-of-the-box in a more pythonic way.

```
def is_unit(unit_string):
    try:
        unyt.Unit(ustr)
    except unyt.exceptions.UnitParseError:
        return False
    return True
```

ngoldbaum:
Closes #30 

I've also removed a yt-specific error message with the intention of adding special logic to catch and report that error in a version of yt that depends on unyt.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -732,10 +732,15 @@ Sometimes it is convenient to create a unit registry containing new units that a
   >>> reg = UnitRegistry()
   >>> reg.add("code_length", base_value=10.0, dimensions=length,
   ...         tex_repr=r"\rm{Code Length}")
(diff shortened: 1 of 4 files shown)


ngoldbaum:
Thanks for giving this a look :)

Issue #27, opened 2018-07-13
Optimize importing unyt


ngoldbaum:
As reported by @majj, importing unyt takes about one second. I suspect there isn't much we can do about this since we need to import both numpy and sympy.

neutrinoceros:
I've been digging into this since the problem has apparently gotten worse.
Here's a graphical profile for `import unyt` (as from the latest master branch)

![Screen Shot 2021-02-08 at 22 55 44](https://user-images.githubusercontent.com/14075922/107286400-d81b3b00-6a60-11eb-86fa-0c6b423c00f4.png)

for reproduction:

```bash
$ python -X importtime import_unyt.py 2> import_unyt.log
$ tuna import_unyt.log
```
with
```python
# import_unyt.py
import unyt
```
and using https://github.com/nschloe/tuna

So indeed, numpy and sympy are taking about a second to import (~0.75s in this particular run), but about 40% (0.9s) of the import time is spent importing `unyt.physical_constants`.
I was able to narrow it down to this line:
https://github.com/yt-project/unyt/blob/aaa2b244207c14c5f472214595a00201d105c0ae/unyt/unit_systems.py#L98
- since no exception is actually caught in this try block when importing unyt, I tried removing the try block and keeping the call, which did not lead to any change I could measure with this (admittedly not very robust) profiling method
- removing the call to `Unit.in_base()` altogether leads to an important gain however reduces the total import time by 0.8s (~35%) so there's room for optimisation here I think.
I may report back if I can grasp a better idea of how to improve on this :)

neutrinoceros:
update: we can track it much more closely with
```
python -m cProfile -o log.pstats import_unyt.py
gprof2dot --colour-nodes-by-selftime log.pstats | dot -Tsvg -o out.svg
```
Where the most time-consuming island looks like this:
![Screenshot 2021-05-01 at 23 13 15](https://user-images.githubusercontent.com/14075922/116795112-df3b6d80-aad2-11eb-999d-5f3f55a79087.png)

so it seems that 928533 deepcopies are performed at import time, which can very likely be optimised.

mattwthompson:
`unyt.physical_constants` does seems to be the culprit for some incredibly slow import times I'm coming across (`python -X importtime -c "import unyt" 2> tuna.log && tuna tuna.log`)


![image](https://user-images.githubusercontent.com/7935382/127569547-3798de5e-d0c1-479e-85ae-fc8d0792e56a.png)

[...]

PR #35, opened 2018-07-23
Add numpy and sympy import checks to top-level __init__.py

### LINKED ISSUE #32: numpy version requirement not evaluated at install
I was able to successfully install unyt (using pip) with numpy 1.10 which is below the requirement (1.13). Then errors are raised during runtime.

ngoldbaum:
Closes #32

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,6 +23,7 @@ deps =
     pytest-cov
     pytest-doctestplus
     flake8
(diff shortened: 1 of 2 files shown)


ngoldbaum:
@jzuhone does this look OK to you? I guess in principle this could make #27 harder but I don't really have high hopes for that.

PR #38, opened 2018-07-25
Add Spatial frequency, with spectral equivalence support

ksunden:
I spoke to @ngoldbaum briefly at Scipy2018 about the spectral equivalence and the units my particular sector of spectroscopy uses. In particular we use wavenumbers  (1/cm) extensively.

This PR adds a spatial frequency dimension `1 / (length)`, and then makes the `spectral` equivalence able to convert to and from spatial_freqency units.

I considered adding `wn` as a unit (equivalent to `1/cm`), but concluded that would likely be better suited to my more specific packages. If the maintainers would like to add it, let me know, perfectly willing to submit such a PR.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -450,7 +450,8 @@ for the problem you are working on.
 The :mod:`unyt` library implements the following equivalencies:
 
 * "thermal": conversions between temperature and energy (:math:`E = k_BT`)
-* "spectral": conversions between wavelength, frequency, and energy for photons
+* "spectral": conversions between wavelength, spatial frequency, frequency, and
+  energy for photons
   (:math:`E = h\nu = hc/\lambda`, :math:`c = \lambda\nu`)
 * "mass_energy": conversions between mass and energy (:math:`E = mc^2`)
 * "lorentz": conversions between velocity and Lorentz factor
@@ -471,7 +472,7 @@ array, use the :meth:`unit_array.list_equivalencies
   schwarzschild: mass <-> length
   compton: mass <-> length
   >>> km.list_equivalencies()
-  spectral: length <-> frequency <-> energy
+  spectral: length <-> spatial_frequency <-> frequency <-> energy
   schwarzschild: mass <-> length
   compton: mass <-> length
 
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1022,7 +1022,7 @@ class unyt_array(np.ndarray):
         -------
         >>> from unyt import km
         >>> (1.0*km).list_equivalencies()
-        spectral: length <-> frequency <-> energy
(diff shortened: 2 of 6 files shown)


ngoldbaum:
Hey I'm glad this was easy :)

This looks correct to me but I'd like @jzuhone to take a look as well.

I guess `wn` means wave-number? I think calling `1/cm` wave number might be a tad too domain-specific (as an astronomer that seems a little weird to me).

ksunden:
Yes indeed `wn` means wavenumber. Figured it may be too specific, so I can just add a `unyt.define_unit('wn', 1/unyt.cm)` within my own package when the time comes to replace our homebuilt, not as general unit system.

Wavenumbers are very popular with those who do infrared spectroscopy.
The first thing I tried to get wavenumbers to work was hackily defining it as an energy unit (rather than properly accounting for the dimensions) as it scales linearly with photon energy. That worked, but I do feel this is a better solution as submitted.

The other avenue I considered before diving into the code itself was allowing equivalence translations to take the inverse of *any* requested unit (as the information to get to `length` was already there, and spatial frequency is just it's multiplicative inverse).
Such an approach would have the advantage of being able to get things like temporal period out of the "spectral" equivalence.
That seemed more complicated to implement, and I have not thought through implications for other kinds of equivalencies. But It's an idea that would provide an alternate route. I do still think defining a `spatial_frequency` dimension would be useful in that case, but the translations would be defined differently perhaps.

ksunden:
I may at some point (especially if I want to use this in theoretical applications) wish to add angular frequencies (both spatial and temporal) to this equivalence as well and complete the cube (with additional offshoot to photon energy):

cube

jzuhone:
Currently at O’Hare with my family catching a plane back to Boston, but i’ll look at this tonight.

jzuhone:
Looks good—definitely supportive of adding angular frequencies here.

ngoldbaum:
Regarding wave numbers, John and I were talking privately over slack about this PR and I realized that there might be a natural way to accomodate them in the library. The reason I was hesitant to call `1/cm` `wn` is that to me a wave number is just one over length, not over over centimeters specifically. I'm also not so hot on the name `wn` being built in to the library.

What do you think about saying, in `dimensions.py`:

```
wave_number = spatial_frequency = 1/length
```

That should make it possible to do e.g.:

```
wn = unit_system['wave_number']
```

And it would return a wavenumber unit appropriate for whatever unit system you've chosen to adopt.

That doesn't have to be part of this pull request, but if you'd be interested in doing a followup that implements this I'd definitely be ok with merging.

In fact I think I'm going to merge this pull request right now along with a quick fix for one of the docstring changes (just expanding the latex formula to include nu_bar).

PR #52, opened 2018-09-03
add unyt_array.argsort. Fixes #51

### LINKED ISSUE #51: argsort returns array of zeros
* unyt version: 1.0.7
* Python version: 3.6.6
* Operating System: Mac OS 10.12.6

### Description

The argsort function returns an array of all zeros with units when operating on a `unyt_array`. Beside being all zeros, I would expect the returned array to be unitless.

### What I Did

```
>>> import numpy as np
>>> import unyt
>>> q = unyt.unyt_array(np.random.random(10), "g")
>>> q.argsort()
unyt_array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 'g')
>>> np.argsort(q)
unyt_array([0, 0, 0, 0, 0, 0, 0, 0, 0, 0], 'g')
```

Not sure if this helps, but in `yt`, I get the following:
```
>>> import numpy as np
>>> import yt
>>> q = yt.YTArray(np.random.random(10), "g")
>>> q.argsort()
YTArray([8, 1, 5, 9, 2, 0, 4, 3, 7, 6]) g
```

The values are correct, but the returned array also has units, which I would not expect.

I'm happy to take a deeper look if someone can point me in the right direction.

ngoldbaum:
There's a doctest in the docstring that should suffice as a test. Let's see what the coverage report says. ping @brittonsmith 

It turns out numpy has an internal hook to call obj.argsort if it's defined for an object passed to `np.argsort`. Not sure why this behaves differently in yt.units, in principle this could be backported there.

### DIFF
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,8 +4,6 @@ environment:
     - TOXENV: py27
     - TOXENV: py37
 
-platform: x64
-
 matrix:
   fast_finish: true
 
--- a/appveyor.yml
+++ b/appveyor.yml
@@ -4,6 +4,8 @@ environment:
     - TOXENV: py27
     - TOXENV: py37
 
+platform: x64
+
 matrix:
   fast_finish: true
 
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1099,6 +1099,24 @@ class unyt_array(np.ndarray):
         """
         return np.array(self)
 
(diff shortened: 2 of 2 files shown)


brittonsmith:
Thanks, @ngoldbaum, this looks good to me. It would be great to backport this to yt.units.

PR #57, opened 2019-01-08
blacken the codebase

ngoldbaum:
This formats the codebase with [black](github.com/ambv/black), a formatter for python code. Overall I'm happy with most of the changes it made. There are a few places where it makes things less clear but I think that's worth it for the automation and lack of need to fight about formatting in the future.

I've added a test for black and have ignored a couple of flake8 rules as suggested in the black docs. I've also increased the max line length to 88 characters.

I'm happy to bikeshed on this stuff, however black has few configuration knobs (on purpose) so there's not much I can do to control the formatting output.


MatthewTurk:
I love this!

ngoldbaum:
OK tests are all passing now. @jzuhone would you mind doing the honors if you're ok with everything?

ngoldbaum:
(I dropped python3.4 testing because numpy just dropped support as well)

PR #58, opened 2019-01-22
add automatic unit simplification. closes #49

### LINKED ISSUE #49: Is it possible to get Hz * s to reduce?
* unyt version: 1.0.6
* Python version: 3.6.6
* Operating System: MacOS 10.13.6

### Description

I multiplied Hz and seconds and got a Hz * s quantity.
I expected a result without dimension, as in Km / Km.

### What I Did

```
In [281]: (1*u.Hz) * (1*u.s)
Out[281]: unyt_quantity(1.0, 'Hz*s')

In [282]: (1*u.km) / (1*u.km)
Out[282]: unyt_quantity(1.0, '(dimensionless)')

In [286]: ((1*u.Hz) * (1*u.s)).to(u.dimensionless)
Out[286]: unyt_quantity(1.0, '(dimensionless)')
```

ngoldbaum:
This adds automatic simplification of units in the form of a new `Unit.simplify()` method. It also makes it so that products and quotients of arrays with units will now automatically be simplified.

Caveats:

* I haven't done any performance benchmarking. In principle calling `simplify()` in the `__array_ufunc__` machinery is "free" because it's behind an LRU cache but I need to actually verify that. There's also an extra multiplication on the array that happens if there's any unit cancellation. In practice though most people would be "manually" doing that simplification and therefore incurring the cost of the extra multiplication anyway. All that said I still need to actually verify that the performance impact is at least acceptable.
* My tests are not *super* extensive and this is an invasive change. It's entirely possible there are untested corner cases.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -253,7 +253,8 @@ dimensions of length::
 
 The :class:`Unit <unyt.unit_object.Unit>` class has a :meth:`simplify()
 <unyt.unit_object.Unit.simplify>` method that we can call to create a new unit
-object to represent this dimensionless conversion factor::
+object to that includes the dimensionless ratio ``m/cm`` as a constant
+coefficient::
 
   >>> (m**2/cm).simplify()
   100*m
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -268,11 +268,11 @@ Products and quotients of unit objects will not be simplified unless
 ``simplify()`` is called explicitly. However, products and quotients of arrays
 and quantities will be simplified to make interactive work more intuitive::
 
-  >>> from unyt import Watt, minute, hour
-  >>> power = [20, 40, 80] * Watt / minute
+  >>> from unyt import erg, minute, hour
+  >>> power = [20, 40, 80] * erg / minute
   >>> elapsed_time = 3*hour
   >>> print(power*elapsed_time)
-  [ 3600.  7200. 14400.] W
+  [ 3600.  7200. 14400.] erg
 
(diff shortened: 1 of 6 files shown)


docs/usage.rst
```diff
@@ -83,16 +83,15 @@ end, days::
 
   >>> period = 2*pi*(semimajor_axis**3/(G*Mjup))**0.5
   >>> period
-  unyt_array([2.64196224e-12, 5.30291672e-12, 1.06837107e-11,
-              2.49212918e-11], 'AU**(3/2)*s/m**(3/2)')
+  unyt_array([ 152867.34547843,  306833.60667034,  618173.2944961 ,
```
jzuhone: This is a really good change here in the docs.

ngoldbaum:
I've done some basic performance testing and don't see any significant slowdowns, e.g.:

```
goldbaum at ROUS in ~/Documents/unyt on simplify!
± python -m perf timeit -s "import numpy as np; data1 = np.random.random(1000); data2 = np.random.random(1000); import unyt as u; data1 *= u.s; data2 *= u.Hz" "data1*data2"
.....................
Mean +- std dev: 9.46 us +- 0.54 us

goldbaum at ROUS in ~/Documents/unyt on master!
± python -m perf timeit -s "import numpy as np; data1 = np.random.random(1000); data2 = np.random.random(1000); import unyt as u; data1 *= u.s; data2 *= u.Hz" "data1*data2"
.....................
Mean +- std dev: 9.35 us +- 0.42 us
```

Going to go ahead and merge. Thank you for looking this over!

PR #61, opened 2019-01-24
fix incorrect conversions between inverse velocities and ohm. 

### LINKED ISSUE #60: Unit conversions with time/length are wrong
* unyt version: 1.0.7
* Python version: 3.6.8
* Operating System: macOS 10.14.2

### Description

When I create quantities with units of `time/length` the unit conversions (e.g., `in_mks` or `in_cgs`) always seem to produce a nonsensical quantity with units of ohms. There is no error or warning produced, just incorrect results.

### What I Did

```
>>> import unyt

# This is to show that units of length/time works fine:
>>> a = unyt.unyt_quantity(100, 'm/s')
>>> a.in_cgs()
unyt_quantity(10000., 'cm/s')

# This is wrong:
>>> b = unyt.unyt_quantity(100, 's/m')
>>> b.in_cgs()
unyt_quantity(8.98755179e+13, 'ohm')
>>> b.in_mks()
unyt_quantity(8.98755179e+13, 'ohm')

# Also:
>>> a.in_cgs().in_cgs()
unyt_quantity(10000., 'cm/s')
>>> b.in_cgs().in_cgs()
unyt_quantity(100., 'statohm')
```

ngoldbaum:
closes #60 

@jzuhone if you could look at this that would be awesome.

There is a behavior change (see the changed test) but I think it's worth making the behavior change to avoid the obviously incorrect behavior in #60. In principle we could also fix #60 and avoid the behavior change by just removing the conversion between ohm and statohm from the `em_conversions` dictionary however the exact same problem could also happen for the other CGS EM units, just less often because they have more unusual dimensions that won't come up in a calculation as often. Better to just remove the possibility of the bug.

Thankfully we're about to do a 2.0 release so behavior changes like this are OK if they're worth it (although should be avoided if we possibly can).

### DIFF
--- a/unyt/tests/test_units.py
+++ b/unyt/tests/test_units.py
@@ -616,6 +616,10 @@ def test_code_unit():
     assert u.get_mks_equivalent() != Unit("ohm")
     assert u.get_cgs_equivalent() == Unit("s/cm")
 
+    u = Unit("kC")
+    assert u.get_cgs_equivalent() == Unit("kesu")
+    assert u.get_cgs_equivalent().get_mks_equivalent() == u
+
     UnitSystem(ureg.unit_system_id, "code_length", "kg", "s", registry=ureg)
 
     u = Unit("cm", registry=ureg)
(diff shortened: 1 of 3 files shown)


jzuhone:
This still works ok with prefixed units, say `kC`?

ngoldbaum:
Good point. Now it does.

PR #63, opened 2019-02-08
fix path issues with new pip versions

ngoldbaum:
Since we have a pyproject.toml we opt into PEP 517/518 with the latest release of pip. One side effect is that the directory setup.py is run inside of is no longer on `sys.path`. We want that behavior though because that's how versioneer works.

Also add a gitignore rule for a new file pip is creating. According to the issue in the comment it looks like we'll probably be able to remove it after the next release of pip though.

### DIFF
--- a/.gitignore
+++ b/.gitignore
@@ -24,6 +24,9 @@ wheels/
 *.egg-info/
 .installed.cfg
 *.egg
+# this can probably be removed at some point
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Self-merging, if someone objects later we can always fix then :)

PR #64, opened 2019-02-11
Improve define_unit consistency with a custom registry

qobilidop:
Previously feeding a tuple with a custom registry unit would err, which probably shouldn't. See this example:

```python
from unyt import UnitRegistry, define_unit

reg = UnitRegistry()
define_unit('code_length', (1, 'm'), registry=reg)
define_unit('code_area', (1, 'code_length**2'), registry=reg)
```

It would result in `unyt.exceptions.UnitParseError: Could not find unit symbol 'code_length' in the provided symbols.` without this patch.

### DIFF
--- a/unyt/tests/test_define_unit.py
+++ b/unyt/tests/test_define_unit.py
@@ -1,6 +1,7 @@
 import pytest
 
 from unyt.unit_object import define_unit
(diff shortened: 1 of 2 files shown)


ngoldbaum:
LGTM!

PR #65, opened 2019-02-13
Unit name resolver

ngoldbaum:
The primary thing this pull request does it make it so that unit names with known alternate names automatically get resolved to a canonical unit name. So, for example:

```
>>> Unit('kilometer/second')
km/s
```

Before this would have raised a parsing error because the parser didn't know what `"kilometer"` or `"second"` was.

I was unable to make this change in isolation, so this pull request also includes changes to how the unit names and physical constants in the `unyt.unit_symbols` and `unyt.physical_constants` namespaces get populated. Instead of everything being defined by hand, we instead use some import-time calculation to generate all the alternate unit names from a set of base names and some simple rules (see the `generate_name_alternatives` function for how this works exactly). I've added tests to make sure that all the unit names that get generated are unique. I've also added a new sphinx extension that automatically generates a nice table of all of the units and physical constants:

image

This still needs to be updated to make python2 work but I figured I'd get this pull request out so I'm not sitting on this for even longer.


MatthewTurk:
If this is going to be unyt 2.0, I think it is reasonable to reduce or remove support for Python 2.

Also, this is pretty awesome.

ngoldbaum:
This was a big enough pain to get working on Python2 that I've decided to go ahead and drop Python2 support with this PR. I spoke with @MatthewTurk and @jzuhone offline about this and they both approved.

The other option would be to drop the unicode unit names for angstrom, ohm, and units with a micro prefix, but I really like those and would hate to see that go just to support a Python version that goes EOL at the end of the year.

@langmm does me dropping Python 2 cause you pain for yggdrasil?

ngoldbaum:
OK, tests are all passing now. I realize this is a big PR so please take your time reviewing.

langmm:
This is alright for yggdrasil I think. I can use pint on Python 2 in the meantime (until Python 2 is officially dead and I can stopped talking myself into supporting it).

ngoldbaum:
Thanks for looking this over John!

PR #68, opened 2019-02-27
Tools and docs for authors writing libraries wrapping unyt

ngoldbaum:
This adds two new functions, `add_symbols` and `add_constants` to the public namespace, adds tests, and documents them. It also adds some extra documentation about how library authors can deal with custom unit systems now that #62 is merged. Finally I noticed some minor doc issues (e.g. formatting problems as well as some minor organizational issues) that I've also fixed. I tried to keep my commits atomic so reviewing commit-by-commit might be easier than looking at the PR diff.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -702,18 +702,8 @@ method:
   <Quantity([1 2 3], 'centimeter')>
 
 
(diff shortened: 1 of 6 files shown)


ngoldbaum:
Thanks for looking this over!

PR #82, opened 2019-04-02
Performance improvements

ngoldbaum:
This includes fixes for several performance issues I found integrating unyt with yt.

Big changes are refactoring of parsing into a new `unyt._parsing` private module, which allows me to refactor the `UnitSystem` class to parse unit names and store unit expressions internally, a big performance win since we can avoid repeatedly and expensively parsing unit strings whenever we go through the unit system machinery.

This also includes a small behavior change in the cgs/mks conversion rules to make them slightly more relaxed. This leads to a big performance wins for all unit conversions so I definitely think it's worth it, see the test changes for the extent of the changes.

Last there are a few other more minor or isolated changes that were all clear-cut performance wins and didn't lead to any test changes.

So long as @jzuhone approves of the test changes I'm going to cut another release based on this PR so I can update the yt tests and hopefully substantially improve the performance over there.

### DIFF
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -1891,7 +1891,7 @@ def test_electromagnetic():
     assert_almost_equal(P_cgs.in_cgs(), P_cgs)
     assert_almost_equal(P_mks.in_cgs(), P_cgs)
     assert_almost_equal(P_cgs.in_mks(), P_mks)
-    assert_almost_equal(P_cgs.in_mks(), P_mks)
+    assert_almost_equal(P_mks.in_mks(), P_mks)
 
     V = unyt_quantity(1.0, "statV")
     V_mks = V.in_units("V")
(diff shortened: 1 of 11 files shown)


ngoldbaum:
Tests are passing, this should be ready to review now.

unyt/tests/test_unyt_array.py
```diff
@@ -1891,10 +1888,10 @@ def test_electromagnetic():
     P_cgs = I_cgs * I_cgs * R_cgs
     assert_equal(P_mks.units.dimensions, dimensions.power)
     assert_equal(P_cgs.units.dimensions, dimensions.power)
-    with pytest.raises(UnitsNotReducible):
-        P_cgs.in_cgs()
-    with pytest.raises(UnitsNotReducible):
-        P_mks.in_cgs()
+    assert_almost_equal(P_cgs.in_cgs(), P_cgs)
+    assert_almost_equal(P_mks.in_cgs(), P_cgs)
+    assert_almost_equal(P_cgs.in_mks(), P_mks)
```
jzuhone: It looks like this same test is repeated twice?

ngoldbaum: thanks!

PR #85, opened 2019-05-02
planck units: Correct units of some constants

musoke:
This is tangentially related to the issue I opened earlier.

Working with version 2.2.0 now, there is a different problem: the constants t_pl, E_pl, q_pl, T_pl are all given units of kg rather than their actual units.  The units seem to be correct though.
I have made a fix in this pull request.


ngoldbaum:
Oops! Thanks for the fix :)

PR #72, opened 2019-03-11
make sure unit systems can be defined using unit objects. fixes #71

ngoldbaum:
ping @JBorrow

It looks like this may have used to work by accident. However it seems like a natural thing to want to do so I've explicitly added support for it, including docs and tests.

PR #86, opened 2019-05-18
allow defining unit systems with a quantity

ngoldbaum:
This addresses the crash @jborrow described in the comments on #72.


JBorrow:
Is it worth adding a test that creates the unit system from top-level units as well (even though these should be equivalent)?

ngoldbaum:
I don’t think that would make any difference.

PR #93, opened 2019-06-25
Support loading json unit registries produced by yt.units

ngoldbaum:
This should hopefully unblock https://github.com/yt-project/yt/pull/2219 once this is merged and we cut a new release yt can depend on. Ping @brittonsmith.

While working on this I found a bug in our definition of light year, it looks like when we went from using CGS base values to MKS the correction to the value of light year didn't go in the correct direction, leaving us off by a factor of 10,000 (100 squared).

Note that we need some minor changes on the yt side (which I have staged already but this needs to be merged first) to fully support loading old ytdata outputs.


jzuhone:
I think we can merge this, right?

ngoldbaum:
Yes I’ve been busy with job interviews and had time to come back to this yet.

matthewturk:
Wow, I didn't notice this till just now!  Thanks, @ngoldbaum !  This is great.  And good luck with interviews!

Issue #92, opened 2019-06-21
checking quantities for floating point equality


berceanu:
So far I am using this hack:

```python
from unyt._testing import assert_allclose_units

def allclose_units(actual, desired, rtol=1e-7, atol=0, **kwargs):
    try:
        assert_allclose_units(actual, desired, rtol, atol, **kwargs)
    except AssertionError:
        return False
    return True
```

Is there a better way?

ngoldbaum:
I'd probably just use the logic in `assert_allclose_units` but make it turn False instead of raising assertion errors. You could then refactor `assert_allclose_units` to use this new utility function internally.

```
import numpy as np
def allclose_units(actual, desired, rtol=1e-7, atol=0, **kwargs):
    # Create a copy to ensure this function does not alter input arrays
    act = unyt_array(actual)
    des = unyt_array(desired)

    try:
        des = des.in_units(act.units)
    except (UnitOperationError, UnitConversionError):
        return False
    
    rt = unyt_array(rtol)
    if not rt.units.is_dimensionless:
        raise RuntimeError("Units of rtol (%s) are not " "dimensionless" % rt.units)

    if not isinstance(atol, unyt_array):
        at = unyt_quantity(atol, des.units)

    try:
        at = at.in_units(act.units)
    except UnitOperationError:
        return False

    # units have been validated, so we strip units before calling numpy
    # to avoid spurious errors
    act = act.value
    des = des.value
    rt = rt.value
    at = at.value

    return np.allclose(act, des, rt, at, **kwargs)
```

berceanu:
Would I put this under `_testing` as well?

ngoldbaum:
So `_testing` is a private namespace (maybe it shouldn't be?) so if we want this to be user-visible it should be in a public spot. I think the place that makes the most sense is `unyt.array`, to go along with the other numpy function wrappers we have in there.

Also it's possible that the `__array_function__` protocol will obviate the need for this since we'll be able to override the behavior of `numpy.allclose` in the library.

PR #94, opened 2019-06-28
allclose_units()

berceanu:
Address https://github.com/yt-project/unyt/issues/92


ngoldbaum:
Hey sorry for taking a few days to respond to this.

So this does look good, although it looks like there's some lines you've added that don't have test coverage. New code in unyt needs to have 100% test coverage. In particular it looks like the error branches in `allclose_units` aren't being tested at the moment.

In addition, making `unyt.testing` public is probably a good change, but more work needs to happen to bring the code up to the standard of the other public modules in unyt. In particular, the `process_warning` function probably shouldn't be public, so it should be renamed to `_process_warning`. In addition, you'll need to make sure that both `unyt.testing` and `assert_allclose_units` shows up in the API documentation. I would also make it so that you can do `from unyt import assert_allclose_units` by reimporting `assert_allclose_units` in the top-level `unyt` namespace. Both of these can be accomplished by editing `unyt/__init__.py`, the docstrings of this file are rendered in the API docs page (e.g. https://unyt.readthedocs.io/en/stable/modules/unyt.html), and this file also defines the top-level `unyt` namespace.

To make sure that `unyt.testing` shows up in the built docs you'll need to build the docs locally and double-check that's the case. To build the docs go into the `docs` directory in the repo and do `make html`. You'll need to have sphinx installed for that to work correctly.

It would also probably be worthwhile to mention `assert_allclose_units` in the narrative documentation, particularly the section on using `unyt` as a part of a larger python codebase.

setup.cfg
```diff
@@ -36,7 +36,7 @@ omit =
     versioneer.py
     unyt/_version.py
     unyt/_on_demand_imports.py
-    unyt/_testing.py
```
ngoldbaum: If we're exposing this publicly I don't think we can get away with excluding it from test coverage anymore.

ngoldbaum:
btw if you're not up for making `unyt.testing` public that's cool, just make it private again and open an issue about maybe making it public in the future.

berceanu:
I’ll give it a shot :)

setup.cfg
berceanu: Added `tests/test_unyt_testing.py`.

berceanu:
Done. Care to have another look?

unyt/array.py
```diff
@@ -2280,3 +2281,37 @@ def savetxt(
...
+
+    if not isinstance(atol, unyt_array):
+        at = unyt_quantity(atol, des.units)
+    else:
+        at = atol
+
+    try:
+        at = at.in_units(act.units)
+    except UnitOperationError:
+        return False
```
ngoldbaum: codecov reports that this line is still untested

unyt/tests/test_unyt_testing.py
```diff
@@ -0,0 +1,58 @@
...
+    a1 = unyt_array([1.0, 2.0, 3.0], "cm")
+    a2 = unyt_array([1.0, 2.0, 3.0], "cm")
+    with pytest.raises(UnitConversionError):
+        assert_allclose_units(a1, a2, atol=unyt_quantity(0.0, "kg"))
+
+
+# from unyt import assert_allclose_units, m
+#  actual = [1e-5, 1e-3, 1e-1] * m
+#  desired = actual.to("cm")
+#  assert_allclose_units(actual, desired)
```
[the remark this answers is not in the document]
berceanu: Sorry! Done.

unyt/array.py
berceanu: Yes, I saw that and I was wondering how to hit that exception..

unyt/__init__.py
```diff
@@ -89,6 +90,7 @@
 from unyt.unit_object import Unit, define_unit  # NOQA: F401
 from unyt.unit_registry import UnitRegistry  # NOQA: F401
 from unyt.unit_systems import UnitSystem  # NOQA: F401
+from unyt.testing import assert_allclose_units  # NOQA: F401
```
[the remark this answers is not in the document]
berceanu: Done.

unyt/array.py
```diff
@@ -2280,3 +2281,37 @@ def savetxt(
         newline="\n",
         comments=comments,
     )
+
+
+def allclose_units(actual, desired, rtol=1e-7, atol=0, **kwargs):
```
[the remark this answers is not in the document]
berceanu: Should it contain example usage and/or tests?

berceanu:
> For the docstring, if you can please include an example. Look at the other docstrings in the file for templates.

Do you mean the docstring for `assert_allclose_units` or for `allclose_units`?

unyt/array.py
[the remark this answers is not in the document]
ngoldbaum: Yup, the example and test are the same thing, the examples get executed during the tests and we make sure the output always stays the same.

unyt/array.py
ngoldbaum: Something like this, I think:

```
with pytest.raises(UnitOperationError):
    allclose_units([1, 2, 3]*kg, [1, 2, 3]*kg, atol=3*meter)
```

(haven't actually tried running that, please double-check me)

ngoldbaum:
I mean the new docstring you'll be adding for `allclose_units`. If `assert_allclose_units` doesn't have an example, it probably should as well :)

berceanu:
All done.

ngoldbaum:
Awesome, thanks so much for sticking with this and carrying it to completion. I hope this builds your confidence to contribute and I get to review more code from you :)

PR #102, opened 2019-08-21
add support for the clip ufunc and add testing for np.clip, closes #101

### LINKED ISSUE #101: numpy.clip is now a ufunc which unit doesn't know about
* unyt version: 2.3.0
* Python version: 3.7
* NumPy version: 1.17
* Operating System: macOS

### Description

In NumPy 1.17 they made `numpy.clip` a ufunc which has 3 inputs. Currently, we do not support any ufuncs with 3 inputs. 

Currently this breaks volume rendering tasks which use sigma clipping in yt if NumPy 1.17 is installed. See https://github.com/yt-project/yt/issues/2313

### Code for reproduction

```python
import yt
import numpy as np
a = yt.YTArray(np.random.uniform(size=10), "cm")
a.clip(0.4,  0.6)
```

gives:

```pycon
RuntimeError                              Traceback (most recent call last)
<ipython-input-1-335caf0515a5> in <module>
      2 import numpy as np
      3 a = yt.YTArray(np.random.uniform(size=10), "cm")
----> 4 a.clip(0.4,  0.6)

~/miniconda3/lib/python3.7/site-packages/numpy/core/_methods.py in _clip(a, min, max, out, casting, **kwargs)
    130     else:
    131         return _clip_dep_invoke_with_casting(
--> 132             um.clip, a, min, max, out=out, casting=casting, **kwargs)
    133
    134 def _mean(a, axis=None, dtype=None, out=None, keepdims=False):

~/miniconda3/lib/python3.7/site-packages/numpy/core/_methods.py in _clip_dep_invoke_with_casting(ufunc, out, casting, *args, **kwargs)
     83     # try to deal with broken casting rules
     84     try:
---> 85         return ufunc(*args, out=out, **kwargs)
     86     except _exceptions._UFuncOutputCastingError as e:
     87         # Numpy 1.17.0, 2019-02-24

~/Source/unyt/unyt/array.py in __array_ufunc__(self, ufunc, method, *inputs, **kwargs)
   1707             raise RuntimeError(
   1708                 "Support for the %s ufunc with %i inputs has not been"
-> 1709                 "added to unyt_array." % (str(ufunc), len(inputs))
   1710             )
   1711         if unit is None:

RuntimeError: Support for the <ufunc 'clip'> ufunc with 3 inputs has not been added to unyt_array.
```

ngoldbaum:
This should probably get backported to yt.units, ping @matthewturk.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -12,7 +12,7 @@ setenv =
     PYTHONPATH = {toxinidir}
 depends = begin
 deps =
-    pytest
+    pytest<5.1
     sympy
     numpy
     h5py
@@ -30,7 +30,7 @@ commands =
 
 [testenv:py35-versions]
 deps =
-    pytest
+    pytest<5.1
     sympy==1.2
     numpy==1.13.3
     h5py==2.6.0
@@ -46,7 +46,7 @@ commands =
 
 [testenv:py35-dependencies]
 deps =
-    pytest
+    pytest<5.1
     sympy
     numpy
     coverage
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -105,7 +105,12 @@ from numpy import (
     ones_like,
     matmul,
 )
-from numpy.core.umath import _ones_like, clip
+from numpy.core.umath import _ones_like
+
+try:
+    from numpy.core.umath import clip
+except ImportError:
+    clip = None
 from sympy import Rational
 import warnings
 
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -105,7 +105,7 @@ from numpy import (
     ones_like,
     matmul,
 )
-from numpy.core.umath import _ones_like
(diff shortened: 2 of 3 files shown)


ngoldbaum:
Looks like tests are passing @jzuhone can you take a look at this?

The changes to the tox config are to work around a compatibility issue between `pytest-doctest-plus` and the latest version of `pytest`. It's fixed upstream but there needs to be a new release for `pytest-doctest-plus` to be able to get it so for now I'm just declaring that we want to use pytest before version 5.1. We can remove that hack once `pytest-doctest-plus` gets a release.

jzuhone:
I'm happy with this. Should we just merge?

ngoldbaum:
Sure, I can do a release this afternoon.

PR #106, opened 2019-09-11
Add support for unit percent

thisch:


thisch:
@ngoldbaum, can you please help me prettify `str()` of a percentage? I would like to see the percent sign `%` in the output.

thisch:
I fixed the issue with the prettified str() output by defining "%" as the `canonical_name` and "percent"  as an alias.

ngoldbaum:
Hey I just pushed to your branch, hope you don't mind! If you prefer I not do that in the future without asking just let me know.

For the pylint comment you added, we don't use pylint, so I removed it. If you want to add pylint checking to this repo then you'd need to touch CI and whatnot. That's totally in-scope but I'd prefer not to add stuff to support pylint without actually fully supporting pylint yet.

It sounds like you figured out how to make `%` the canonical name but perhaps you didn't push that commit? In any case I went ahead and did that as well.

Finally I think just using exact string comparisons in the tests is simplest and I don't expect those exact strings to ever change so I've removed the use of `re` you added.

I'll give you a few days to respond then merge this if I don't hear anything, assuming the tests all pass.

Thanks very much for the contribution and sorry for taking a couple of weeks to look at it. I just started a new job this week where I get to work on OSS development 25% of my time, so I'm taking advantage of that now :)

thisch:
You are right, I forgot to push my latest changes. Thx for fixing it!

Regarding using `re`.  The regex code is only needed for making travis CI happy.

> Thanks very much for the contribution and sorry for taking a couple of weeks to look at it.

Never mind!

>  I just started a new job this week where I get to work on OSS development 25% of my time, so I'm taking advantage of that now :)

Great! I would love to have such a job ;) I'll use this change and open a new issue about performance improvements ;)

ngoldbaum:
> Regarding using re. The regex code is only needed for making travis CI happy.

Ah I see, because string printing changed in newer numpy versions.

I think I have an alternate fix. Eventually we'll drop support for old numpy versions that have the legacy printing behavior and avoid these annoyances..

PR #108, opened 2019-09-23
Adding a test that fails for yt on appveyor.

matthewturk:
The clip function in yt master is causing issues when called from within the ytdata frontend.  This test seems to be the smallest way I can reproduce it.  However, I can only make it fail on appveyor.  I'm optimistic that issuing this pull-request will help track down why it's problematic, and where the (platform specific?) failure occurs.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1712,7 +1712,12 @@ class unyt_array(np.ndarray):
                     )
         else:
             if ufunc is clip:
-                inp = (inputs[0].view(np.ndarray), inputs[1], inputs[2])
+                inp = []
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Looks like the way I implemented clip wasn't covering all the edge cases. I'll try to take a look today. Would you mind if I pushed to this branch?

matthewturk:
That would be great, yes -- thank you!

ngoldbaum:
My guess is that the reason you're only seeing this on appveyor is that the other test suites are installing and older version of numpy.

PR #109, opened 2019-10-04
Allow == and != for data with different units. Fixes #105.

### LINKED ISSUE #105: numpy DeprecationWarning when quantities with different units are compared.
* unyt version:
2.3.1
* Python version:
3.7
* Operating System:
GNU/Linux & conda env

### Description

We noticed that unyt outputs a deprecationwarning when quantities with different units are compared. 
### What I Did

```
In [3]: 1*u.mm == 1*u.V                                                                                                                                      
/opt/miniconda/envs/py3/bin/ipython:1: DeprecationWarning: elementwise comparison failed; this will raise an error in the future.
  #!/opt/miniconda/envs/py3/bin/python
```
Note that no DeprecationWarning is output when the units are rescalable:
```

In [5]: 1*u.mm == 1*u.m                                                                                                                                      
Out[5]: array(False)
```

ngoldbaum:
Up until now we were unconditionally raising a UnitOperationError for all
comparisons of data with different units, including `equal` and `not_equal`. Due
to details of how NumPy handles this internally (I guess it happens in C code
because I don't see any of this when stepping through in pdb) numpy swallows
that error and raises a DeprecationWarning (!?) causing confusing behavior when
comparing quantities with units that are not dimensionally equivalent.

The fix is to add special handling for equal and not_equal when comparing data
with units that have different dimensions. I've also added tests for this case.

This may have negative performance implications for some workflows however I
think we have to do this in the interest of correctness. Also comparing data
with different units like this is kinda a weird thing to do and hopefully won't
impact common workflows.

Thanks to @thisch for pointing this out.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1673,7 +1673,26 @@ class unyt_array(np.ndarray):
                             elif u1.is_dimensionless:
                                 u1 = u0
                             else:
-                                raise UnitOperationError(ufunc, u0, u1)
+                                # comparison with different units, so need to check if
+                                # this is == and != which we allow and handle in a
+                                # special way using an early return from __array_ufunc__
+                                if ufunc in (equal, not_equal):
+                                    if ufunc is equal:
+                                        func = np.zeros_like
+                                    else:
+                                        func = np.ones_like
+                                    ret = func(np.asarray(inp1), dtype=bool)
+                                    if out is not None:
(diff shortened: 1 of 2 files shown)


jzuhone:
Looks fine to me!

thisch:
Thx for the fix @ngoldbaum !

PR #111, opened 2019-10-11
simplify creating arrays and quantities with units

ngoldbaum:
Before:

```
In [2]: %timeit 3 * u.g
4.6 µs ± 56.2 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

After:

```
In [3]: %timeit 3 * u.g
2.94 µs ± 47.6 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

So about 35% faster!

I'm getting rid of `UnitDtypeError` because it wasn't tested beyond the
docstring for `UnitDtypeError` itself and `UnitOperationError` is just as
easy to understand.

I also added `__slots__` both to speed up attribute access as well as to improve
memory footprint.

`ImportCache` is needed because the caching trick won't work on the class itself because `__slots__` is defined so now I just create a singleton to hold the cache.

ping @thisch

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1952,7 +1952,10 @@ class unyt_quantity(unyt_array):
             )
         if units is not None:
             input_units = units
-        if not isinstance(input_scalar, (numeric_type, np.number, np.ndarray)):
+        if not (
+            bypass_validation
+            or isinstance(input_scalar, (numeric_type, np.number, np.ndarray))
+        ):
             raise RuntimeError("unyt_quantity values must be numeric")
         if input_units is None:
             units = getattr(input_scalar, "units", None)
--- a/unyt/exceptions.py
+++ b/unyt/exceptions.py
@@ -301,29 +301,3 @@ class IllDefinedUnitSystem(Exception):
             "Cannot create unit system with inconsistent mapping from "
             "dimensions to units. Received:\n%s" % self.units_map
         )
-
-
-class UnitDtypeError(Exception):
-    """Raised when applying units to invalid data
-
-    Example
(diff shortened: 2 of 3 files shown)


ngoldbaum:
I messed up the performance comparison I originally had on this pull request, I've updated the description with the actual numbers (which are pretty great!).

ngoldbaum:
@jzuhone any chance you can give this a look?

thisch:
Thx a lot for working on this!  I don't get the same performance speedup as you but only 25%.

```
In [3]: %timeit 3 * unyt.mm                                                                                                                                                            
4.79 µs ± 28.7 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)

```

after

```
In [2]: %timeit 3 * unyt.mm                                                                                                                                                            
3.64 µs ± 65 ns per loop (mean ± std. dev. of 7 runs, 100000 loops each)
```

but still a good improvement!

ngoldbaum:
Yes, it likely depends on lots of factors, e.g. do you have fastcache installed, sympy, numpy, and python version etc etc.

thisch:
> Yes, it likely depends on lots of factors, e.g. do you have fastcache installed, sympy, numpy, and python version etc etc.

No, I didn't have fastcache installed. Now I have, but the results are basically the same. 

```
fastcache                 1.1.0            py37h516909a_0    conda-forge
numpy                     1.17.2           py37h95a1406_0    conda-forge
python                    3.7.3                h33d41f4_1    conda-forge
sympy                     1.4                      pypi_0    pypi
```

PR #114, opened 2019-12-11
Fix unit hash function

saethlin:
I discovered that the `__hash__` implementation for `Unit` is incorrect when looking at a perf bottleneck in some code; things were slowing down because the `@lru_cache` on `_multiply_units` was very rarely being hit.

The Python docs say:
> Objects which are instances of user-defined classes are hashable by default. They all compare unequal (except with themselves), and their hash value is derived from their id().

The previous implementation which just forwarded to the default hash would vary based on the `id()` of the object, and thus the `@lru_cache` would always miss.

I'm not entirely sure this implementation is correct. This is what I invented late last night, but it looks a bit unlike the `__eq__`, but my guess at a test cast that would prove this incorrect isn't failing.

### DIFF
--- a/unyt/tests/test_units.py
+++ b/unyt/tests/test_units.py
@@ -791,3 +791,10 @@ def test_percent():
 
     assert a == c
     assert c == d
+
+
+def test_equal_has_same_hash():
+    a = Unit("m")
+    b = Unit("m")
+    assert a == b
+    assert hash(a) == hash(b)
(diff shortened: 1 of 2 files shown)


ngoldbaum:
Sorry for taking so long to review this. I think this is correct and thank you for catching this error and fixing it.

ngoldbaum:
Hmm I think this actually isn’t quite right. Right now we’re passing `hash` a string object and getting a hash value based on the properties of that object rather than any data actually in the unit system. It would be better to convert the data in the string, which looking at the code is an md5 hash digest, into an integer by parsing the hex digits and xoring with that integer instead. Just commenting here so I don’t forget about this tomorrow, not expecting you to implement it.

PR #115, opened 2019-12-31
Add support for latest versions of h5py, sympy, numpy, and pytest 

ngoldbaum:
This fixes a number of issues caused by updates to upstream libraries we depend on. This is a bit of a grab-bag PR that makes the tests pass again on the latest versions of our dependencies. Unfortunately I can't easily issue these changes as separate PRs since the tests will be broken without all of these coming in at once. If you have questions about reviewing this I'm happy to help out.

To catch issues caused by upstream changes more quickly in the future I've updated the tox config to always recreate a virtualenv. This will ensure we get test failures on the cron job test whenever an upstream release breaks our test suite.

I also made the following updates to our tox config:

* Add testing for python 3.8, update older python minor versions
* Remove pytest version constraint now that issue with pytest-doctest-plus has been fixed.
* Pin `coverage` to be older than coverage 5.0, which allows us to work around https://github.com/nedbat/coveragepy/issues/883

There also some code changes, in particular

* Avoid deprecation warnings from h5py by always specifying the file mode in `h5py.open`.
* Fix usage of `rationalize` transformer in our usage of `sympy.parse_expr`. Before we were not using it correctly since we needed to run `auto_number` first.
* Fix latex output for non-float unit scalar multiples.
* Add support for NumPy 1.8 by fixing handling of the `out` keyword argument for ufuncs that have multiple outputs.
* Fix warnings from Pint caused by our implicity using `Pint.quantity.__array__`.
* Make the unit registry saved to HDF5 files smaller to avoid the attribute size limit triggered in newer versions of h5py. Fixes https://github.com/yt-project/yt/issues/2404.

### DIFF
--- a/.travis.yml
+++ b/.travis.yml
@@ -29,10 +29,11 @@ before_install:
   - export PYENV_ROOT="$HOME/.pyenv"
   - export PATH="$HOME/.pyenv/bin:$PATH"
   - eval "$(pyenv init -)"
-  - pyenv install -s 3.5.5
-  - pyenv install -s 3.6.5
-  - pyenv install -s 3.7.1
-  - pyenv local 3.5.5 3.6.5 3.7.1
+  - pyenv install -s 3.5.7
+  - pyenv install -s 3.6.9
+  - pyenv install -s 3.7.5
+  - pyenv install -s 3.8.0
+  - pyenv local 3.5.7 3.6.9 3.7.5 3.8.0
   - pip install tox tox-pyenv codecov twine
 
 # Command to run tests, e.g. python setup.py test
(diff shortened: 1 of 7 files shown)


jzuhone:
> Add support for NumPy 1.8 by fixing handling of the out keyword argument for ufuncs that have multiple outputs.

Do you mean NumPy 1.18?

ngoldbaum:
Yes, sorry for the typo.

jzuhone:
@ngoldbaum do you want anyone else to have a look at this before merging?

ngoldbaum:
I was a little worried about the test failures in yt from the changes in this PR. I'm not convinced that the yt failures are something that can't happen with unyt, since unyt's `Unit` class doesn't subclass sympy's `Expr` class. I'll go ahead and merge this in now. Thanks for taking a look John.

PR #56, opened 2018-12-18
Support integer and reduced precision arrays. Fixes #55

ngoldbaum:
Up to now we assumed pretty much everywhere that data are 64 bit floats or that data will eventually get converted to 64 bit floats. This led to weird behavior if people tried to use integer or reduced precision floats, see e.g. the array printing weirdness reported in #55.

My justification for not supporting ints was that data with units are "naturally" floats. However sometimes it's important for e.g. memory layout reasons to have data as ints but still track unit metadata. Perhaps more importantly for memory layout or memory size reasons it's also important to store data in 32 bit or 16 bit precision if that's OK for the kind of data that a person is working with.

This PR changes that behavior and makes it so that we infer the dtype of data using the same rules as `np.array` for data where no dtype is explicitly specified.

We also try to keep data the same dtype but will convert from int to float to convert data to a different unit. This can lead to mutation of the dtype of an array if you do an in-place conversion which might be confusing but I think is more useful than either truncating or raising an error. See the docs I added for more detail.

This is a pretty big change. I think if this gets merged we should do a release and call it version 2.0.

PR #119, opened 2020-01-13
coerce data to the destination dtype before multiplying by any converion factors

### LINKED ISSUE #118: Equality test of equivalent quantities, but with different prefixes, returns False.
* unyt version: 2.4.1
* Python version: 3.8.1
* Operating System: Win10

### Description

The quantities 1 s and 1000 ms are equal, but unyt says they're not equal.

### What I Did

```
>>> from unyt import s, ms
>>> 1*s == 1000*ms
array(False)
```
I also find the rather surprising result:
```
>>> 1*s >= 999*ms
array(True)
>>> 1*s >= 1000*ms
array(False)

ngoldbaum:
fixes #118

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1728,7 +1728,7 @@ class unyt_array(np.ndarray):
                             "cannot by multiplied, divided, subtracted or "
                             "added with data that has different units."
                         )
-                    inp1 = np.asarray(inp1) * conv
+                    inp1 = np.asarray(inp1, dtype=new_dtype) * conv
             # get the unit of the result
             mul, unit = unit_operator(u0, u1)
             # actually evaluate the ufunc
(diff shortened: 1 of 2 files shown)


l-johnston:
While I agree that this could fix it, fundamentally the problem is in line 1732 of array.py:
new_dtype = np.dtype("f" + str(inp1.dtype.itemsize))

In this case, inp1 is 1000*ms and inp1.dtype is int32 on Win10 and int64 on Linux. Win10 floats are float64.

On Win10: 1.0*s == 1000.0*ms is True.

Would it be better to promote int32 to int64?

ngoldbaum:
No, because I don't want to force things to have a wider size than what was input originally. We were handed 32 bit data, the user shouldn't get back 64 bit data unless they explicitly ask for it. This is an explicit design decision in unyt, see https://github.com/yt-project/unyt/pull/56.

PR #123, opened 2020-01-19
Try to speed up UnitRegistry.from_json

saethlin:
We have a code that does some analysis on snapshots then dumps HDF5 files which serialize a bunch of unyt_arrays and unyt_quantities. It also serializes the unit system that was created when the snapshot was opened. So when someone wants to loop over hundreds of these files which contain serialized unyt structures, the process overall is dominated by reconstituting the serialized unit registry for each file.

The `functools.lru_cache` can't be keyed on both the arguments to the `sympify` call, because the second argument is not hashable. I think this is okay because modifying the module seems like it would invalidate other things as well.

### DIFF
--- a/unyt/unit_registry.py
+++ b/unyt/unit_registry.py
@@ -14,6 +14,7 @@ A registry for units that can be added to and modified.
 
 
 import json
+from functools import lru_cache
 
 from unyt import dimensions as unyt_dims
 from unyt.exceptions import SymbolNotFoundError, UnitParseError
@@ -37,6 +38,15 @@ def _sanitize_unit_system(unit_system, obj):
     elif unit_system == "code":
         unit_system = obj.units.registry.unit_system_id
     return unit_system_registry[str(unit_system)]
+
+
+@lru_cache
+def cached_sympify(u):
+    """
+    Successive loads of unit systems produce the same calls to sympify in UnitRegistry.from_json
+    Even within a single load, this is a net gain because there will often be a few cache hits
+    """
+    return sympify(u, locals=vars(unyt_dims))
 
 
 class UnitRegistry:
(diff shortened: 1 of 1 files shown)


unyt/unit_registry.py
```diff
@@ -39,6 +40,15 @@ def _sanitize_unit_system(unit_system, obj):
     return unit_system_registry[str(unit_system)]
 
 
+@lru_cache
```
ngoldbaum: To support Python3.7 and older, this needs to be `@lru_cache()` see https://stackoverflow.com/questions/47218313/use-functools-lru-cache-without-specifying-maxsize-parameter

unyt/unit_registry.py
```diff
@@ -39,6 +40,15 @@ def _sanitize_unit_system(unit_system, obj):
...
+def cached_sympify(u):
+    """
+    Successive loads of unit systems produce the same calls to sympify in UnitRegistry.from_json
+    Even within a single load, this is a net gain because there will often be a few cache hits
```
ngoldbaum: `flake8` is complaining that these lines are too long. In general python docstrings should be less than 72 characters wide so they fit comfortable in an 80 character wide terminal.

unyt/unit_registry.py
saethlin: In the rest of unyt, I see `@lru_cache(maxsize=128, typed=False)`. Those are the default parameters, but the fact that they're repeated in the code made me think there was some thought put into the choice. Was there? Should there be other values here?

ngoldbaum: No, I just wanted to make it explicit in case I ever wanted to experiment with changing them.

PR #126, opened 2020-01-21
Update unyt_arrayConverter to handle sequences of unyt types.

l-johnston:


unyt/mpl_interface.py
```diff
@@ -98,11 +99,39 @@ def convert(value, unit, axis):
...
+        def enable(self):
+            self.__enter__()
+
+        def disable(self):
+            self.__exit__(None, None, None)
 
-    registry[unyt_array] = unyt_arrayConverter()
-    registry[unyt_quantity] = unyt_arrayConverter()
+        def __repr__(self):
+            return "<MplUnitsCM>"
```
ngoldbaum: The custom repr isn’t really helpful in my opinion, I’d just delete it.

unyt/mpl_interface.py
```diff
@@ -98,11 +99,39 @@ def convert(value, unit, axis):
...
+                value_type = type(value)
+                converted_value = []
+                for obj in value:
+                    converted_value.append(obj.to(*unit))
+                converted_value = value_type(converted_value)
+            else:
+                raise ConversionError(f"unable to convert {value}")
+            return converted_value
+
+    class MplUnitsCM:
```
ngoldbaum: bikeshedding, but can this name be changed? Maybe it could be called `unyt_plotting` or `matplotlib_support`. For the second option people can do (and you can write it in the docs as) `with unyt.matplotlib_support`, which I think is easier to remember than the current name. I realize that pep8 says class names should be camel cased but I think in places that classes get exposed in a top-level API like this it’s ok to use lowercase underscore names since they’re easier to type and remember

unyt/mpl_interface.py
```diff
@@ -98,11 +99,39 @@ def convert(value, unit, axis):
...
+            if isinstance(value, (unyt_array, unyt_quantity)):
+                converted_value = value.to(*unit)
+            elif isinstance(value, (list, tuple)):
+                value_type = type(value)
+                converted_value = []
+                for obj in value:
+                    converted_value.append(obj.to(*unit))
+                converted_value = value_type(converted_value)
+            else:
+                raise ConversionError(f"unable to convert {value}")
```
ngoldbaum: f-strings aren’t available on older versions of python

unyt/__init__.py
```diff
@@ -96,7 +96,9 @@
 from unyt.unit_systems import UnitSystem  # NOQA: F401
 from unyt.testing import assert_allclose_units  # NOQA: F401
 from unyt.dimensions import accepts, returns  # NOQA: F401
-from unyt.mpl_interface import MplUnitsCM  # NOQA: F401
+from unyt.mpl_interface import matplotlib_support  # NOQA: F401
+
+matplotlib_support = matplotlib_support()
```
ngoldbaum: Won’t this turn on matplotlib support if someone imports unyt?

l-johnston: no.
```python
>>> import unyt
>>> import matplotlib.units as mplunits
>>> mplunits.registry
{}
>>> with matplotlib_support:
        print(mplunits.registry)
{unyt_array...}
>>> mplunits.registry
{}
```

unyt/__init__.py
```diff
@@ -96,7 +96,9 @@
 from unyt.unit_systems import UnitSystem  # NOQA: F401
 from unyt.testing import assert_allclose_units  # NOQA: F401
 from unyt.dimensions import accepts, returns  # NOQA: F401
-from unyt import mpl_interface  # NOQA: F401
+from unyt.mpl_interface import matplotlib_support  # NOQA: F401
```
ngoldbaum: This needs either a try/except here or you need to define a dummy `matplotlib_support` in `mpl_interface` in the try/except there. That will fix the test failures on the jobs where we don’t install matplotlib.

It occurs to me that some docs on how to use pyenv and tox to run the full test suite locally would be helpful :)

l-johnston: Yes!!!!

unyt/tests/test_mpl_interface.py
```diff
@@ -2,8 +2,9 @@
 import numpy as np
 import pytest
 from unyt._on_demand_imports import _matplotlib, NotAModule
-from unyt import s, K
+from unyt import s, K, unyt_array, unyt_quantity, matplotlib_support
```
ngoldbaum: looks like you need to add try/except blocks here too

unyt/mpl_interface.py
```diff
@@ -18,6 +18,7 @@
         ConversionInterface,
         AxisInfo,
         registry,
+        ConversionError,
```
ngoldbaum: It looks like importing this broke support for matplotlib 2.2.4. Can you mock `ConversionError` in the `except ImportError` block below?

ngoldbaum:
There were some finicky documentation formatting issues I fixed in the last commit I just pushed. As soon as the tests pass I'm going to merge this. Thank you again for spending your time today on this.

Issue #128, opened 2020-01-22
Would you be at all interested in adding a `name` field to the `unyt.unyt_array` class?


l-johnston:
Would you be at all interested in adding a `name` field to the `unyt.unyt_array` class? I have implemented in this in a few libraries that use `unyt` and find it very useful. Then you can, on these axes, use the names of the (last) arrays that were plotted to form the axes labels.

...

_Originally posted by @JBorrow in https://github.com/yt-project/unyt/issues/125#issuecomment-576899749_

JBorrow:
Thanks for opening this!

l-johnston:
I have implemented this feature and now the axis label will include the array name:

```python
>>> from unyt import *
>>> import matplotlib.pyplot as plt
>>> matplotlib_support.enable()
>>> matplotlib_support.label_style = "[]"
>>> x = [0,1,2]*s
>>> x.name = "Time"
>>> y = [3,4,5]*m
>>> y.name = "Distance"
>>> plt.plot(x, y)
>>> plt.show()
```
![fig5](https://user-images.githubusercontent.com/17813788/72892865-3dc54980-3cdd-11ea-8a6a-ec357f1c8c3d.png)

JBorrow:
Can you also pass a `name` keyword to the `unyt_array` constructor?

So:
```python
x = unyt.unyt_array([1,2,3], units="kpc", name="Radii")
```

JBorrow:
P.s. this is great!

l-johnston:
> 
> 
> Can you also pass a `name` keyword to the `unyt_array` constructor?
> 
> So:
> 
> ```python
> x = unyt.unyt_array([1,2,3], units="kpc", name="Radii")
> ```

possible, but I'd like to hear from @ngoldbaum first.

PR #129, opened 2020-01-22
add 'name' attribute to unyt_array and propagate to matplotlib axis l…

l-johnston:
…abel #128


unyt/array.py
```diff
@@ -482,6 +482,7 @@ def __new__(
         dtype=None,
         bypass_validation=False,
         input_units=None,
+        name="",
```
JBorrow: Shouldn't this be initialised to `None` not an empty str (as empty strs are mutable?)

unyt/mpl_interface.py
```diff
@@ -47,24 +47,31 @@ def axisinfo(unit, axis):
...
+            name = getattr(axis, "_unytarrayname", "")
             if unit_obj.is_dimensionless:
-                label = ""
+                label = name
             else:
+                name += " "
                 unit_str = unit_obj.latex_representation()
                 if unyt_arrayConverter._labelstyle == "[]":
-                    label = "$\\left[" + unit_str + "\\right]$"
+                    label = name + "$\\left[" + unit_str + "\\right]$"
```
JBorrow: Probably want a space between `name` and unit here.

unyt/array.py
l-johnston: I started with None. I'll take a look.

unyt/mpl_interface.py
l-johnston: there is, the space is added prior to this line

unyt/array.py
```diff
@@ -1581,6 +1582,18 @@ def ua(self):
...
+    def name(self):
+        """Return the name (string) of the array"""
+        name = getattr(self, "_name", "")
+        if name is None:
+            name = ""
+        return name
+
+    @name.setter
+    def name(self, name):
+        self._name = name
```
ngoldbaum: Why does this need to be a property? I think it's simpler just to make this an attribute and delete the custom getter/setter unless you want to do some additional checking in the setter. I guess you could check if the provided name is really a string and if it's not then raise an error.

unyt/mpl_interface.py
```diff
@@ -81,6 +88,7 @@ def default_units(x, axis):
 
             Unit object
             """
+            axis._unytarrayname = x.name
```
ngoldbaum: Can we avoid monkeypatching matplotlib objects? What if on the matplotlib side the `axis` object gets destroyed. We have no control over that so I think it would be better to figure out how to achieve the same thing fully inside of unyt.

unyt/mpl_interface.py
```diff
@@ -164,13 +173,21 @@ def label_style(self, label_style="()"):
...
 
         def disable(self):
-            self.__exit__(None, None, None)
+            if self._enabled:
+                self.__exit__(None, None, None)
+
+        @property
+        def enabled(self):
+            """Return the enable state"""
+            return self._enabled
```
ngoldbaum: Why does this need to be added? Is there something broken about the implementation that got merged last night?

l-johnston: I like to use matplotlib_support as a toggle in an interactive session where I may not remember or be able to see easily which state it is in. I could just call matplotlib_support.enable().

unyt/mpl_interface.py
l-johnston: well, the Axis object passed to unyt_arrayConverter.axisinfo(unit, axis) does not contain or point to the underlying unyt_array for that axis, and the unit object can be just a string passed in from plt.plot(x, y, xunits="ms") for example.

I agree and my original implementation attempted to handle this solely within Unyt. In one approach, I added an ``owner`` attribute to Unit that would allow me to get back to the original unyt_array that has the ``name``, but matplotlib's design breaks this approach.

unyt/array.py
l-johnston: In my original implementation, ``name`` wasn't in the constructor hence the property.

jzuhone:
I agree with @ngoldbaum--I'm a bit iffy on this, for the two reasons he stated:

1. Plotting doesn't seem to be enough of a use case to justify adding names to `unyt_array` objects when one just needs to set the appropriate label in Matplotlib. I'd prefer to see more reasons why we'd use name attributes before approving. 
2. Names don't get passed through operations. The problem is that often we don't want that to happen, but sometimes we might. And the boundary between those two situations is somewhat arbitrary and would be hard to code up.

unyt/mpl_interface.py
jzuhone: I share @ngoldbaum's worry here.

l-johnston:
> 
> 
> I agree with @ngoldbaum--I'm a bit iffy on this, for the two reasons he stated:
> 
>     1. Plotting doesn't seem to be enough of a use case to justify adding names to `unyt_array` objects when one just needs to set the appropriate label in Matplotlib. I'd prefer to see more reasons why we'd use name attributes before approving.
> 

Pandas has a name attribute. As is, Pandas will convert unyt_array to ndarray stripping the units. Without implementing their extension interface, a user could propagate useful info. Here's an example.
```python
>>> import pandas as pd
>>> from unyt import s
>>> t = [0, 1, 2]*s
>>> t.name = "time"
>>> t_s = pd.Series(t, name=t.name+f" ({t.units})")
>>> t_s
0    0
1    1
2    2
Name: time (s), dtype: int64
```

unyt/mpl_interface.py
```diff
@@ -18,12 +18,14 @@
...
     pass
 else:
+    from weakref import WeakKeyDictionary
     from unyt import unyt_array, unyt_quantity, Unit
 
     class unyt_arrayConverter(ConversionInterface):
         """Matplotlib interface for unyt_array"""
 
         _labelstyle = "()"
+        _axisnames = WeakKeyDictionary()
```
ngoldbaum: What if someone makes a multipanel plot with two x-axes. Won't the second x-axis clobber the first one?

In general having mutable state defined at class level like this is usually not what you want. Since this class normally acts as a singleton that's probably actually OK here, but then you need to address that this is a global key-value store and find a way to make the keys unique. Right now I don't see how you're enforcing uniqueness in the keys.

l-johnston: > What if someone makes a multipanel plot with two x-axes. Won't the second x-axis clobber the first one?

Matplotlib will generate two unique XAxis instances that are passed to our conversion interface. So, it won't clobber the label. I added a test for multiple subplots.



> Right now I don't see how you're enforcing uniqueness in the keys.

The Axis object is hashable. I have added a test for this. Is hashable sufficient?

ngoldbaum: Oops, thought you were using the name, sorry for misreading.

I'm still torn on this and want to wait until next week to come to a decision.

ngoldbaum:
OK, after taking a look at the examples in your last comment I’m coming around to merging this.

Before merging this there needs to be some discussion about this feature in the narrative docs. You also need to add an entry for `name` in the docstrings for `unyt_array` and `unyt_quantity`. It would also be good to get another plotting example (with an image) that shows how the `name` influences the plot labels.

docs/usage.rst
```diff
@@ -1182,6 +1182,31 @@ function. In other words, apply or check units at interfaces, but during an
...
+  >>> dt = [(a.name, "O") for a in data]
+  >>> data_points = np.array(list(zip(*data)), dtype=dt).view(np.recarray)
+  >>> data_points[0].time
+  unyt_quantity(0, 's')
+  >>> data_points[0].distance
+  unyt_quantity(3, 'm')
+
+.. note::
+  The name attribute does not propagate through mathematical operations or even the
+  unyt_array.copy() function. The only exception to this is the unyt_array.convert_to_units()
```
ngoldbaum: Now that you mention it, I would actually make `copy` propagate that attribute, otherwise it might surprisingly not be a copy in just this one way.

ngoldbaum: And maybe `in_units` and `to` as well?

docs/usage.rst
```diff
@@ -1194,9 +1219,9 @@ Matplotlib is Unyt aware. After enabling support in :mod:`unyt` using the
 manager, Matplotlib will label the x and y axes with the units.
 
   >>> import matplotlib.pyplot as plt
-  >>> from unyt import s, K, matplotlib_support
-  >>> x = [0.0, 0.01, 0.02]*s
-  >>> y = [298.15, 308.15, 318.15]*K
+  >>> from unyt import matplotlib_support, unyt_array
+  >>> x = unyt_array([0.0, 60.0, 120.0], "s")
+  >>> y = unyt_array([298.15, 308.15, 318.15], "K")
```
ngoldbaum: Why change how these are initialized? I think it’s more idiomatic to create a `unyt_array` via multiplication and I’d prefer it if we emphasized that in the docs.

unyt/mpl_interface.py
```diff
@@ -124,23 +148,16 @@ def convert(value, unit, axis):
...
-        ----------
+        """Context manager for enabling the feature
 
-        label_style : str
-          One of the following set, ``{'()', '[]', '/'}``. These choices
-          correspond to the following unit labels:
-
-            * ``'()'`` -> ``'(unit)'``
-            * ``'[]'`` -> ``'[unit]'``
-            * ``'/'`` -> ``'q_x / unit'``
```
ngoldbaum: Why remove the docs for `label_style`? If possible I’d like to keep them even if it duplicates the entry on the property below.

unyt/mpl_interface.py
```diff
@@ -164,13 +181,19 @@ def label_style(self, label_style="()"):
...
         def enable(self):
             self.__enter__()
 
         def disable(self):
-            self.__exit__(None, None, None)
+            if self._enabled:
+                self.__exit__(None, None, None)
+
+
+__all__ = ["matplotlib_support"]
```
ngoldbaum: Can you move this to the top of the file, below the imports? I think that’s more idiomatic and discoverable.

unyt/mpl_interface.py
l-johnston: the duplication was distracting when viewing the api documentation page - it's a tradeoff. I'll add it back.

docs/usage.rst
l-johnston: ``to`` and ``in_units`` create a unyt_array that is missing the name attribute.
```python
>>> from unyt import s
>>> x = [0,1,2]*s
>>> hasattr(x, "name")
True
>>> y = x.to("ms")
>>> hasattr(y, "name")
False
```
Should I investigate?

l-johnston: I updated ``copy`` to propagate ``name``. now,
```python
>>> from unyt import s
>>> x = [0,1,2]*s
>>> x.name = "time"
>>> y.copy()
>>> y.name
'time'
```

ngoldbaum: I bet it has something to do with the way the returned array is created: https://github.com/yt-project/unyt/blob/595b9cdad094a9cd7f04980aee13a96344c0fc5f/unyt/array.py#L819

If I had to hazard a guess, `bypass_validation=True` makes it so you return from `__new__` before `name` gets set. Maybe `name` needs to get set earlier? `bypass_validation` is there for performance, in code where you’re already sure that what you’re passing in is an `ndarray` and a `Unit` object, you can set that to `True` and bypass a bunch of validation code which can substantially improve performance in hot code paths. It’s mostly an internal hack that enables some of the speed of `unyt`.

l-johnston: correct. now ``to`` and ``in_units`` have the attribute. Should I propagate name for these two functions?

ngoldbaum: Yes I think so. Just changing the units doesn’t really change the semantic meaning of the data. Doing a mathematical operation might, so we don’t propagate `name` in `__array_ufunc__`.

unyt/array.py
```diff
@@ -986,7 +994,7 @@ def in_mks(self):
 
     def convert_to_equivalent(self, unit, equivalence, **kwargs):
         """
-        Return a copy of the unyt_array in the units specified units, assuming
```
ngoldbaum: good catch, thank you!

jzuhone:
Other than @JBorrow's latest comments, this looks good to me.

unyt/array.py
```diff
@@ -355,6 +355,9 @@ class unyt_array(np.ndarray):
         corrupted, invalid units or array data, but can lead to significant
         speedups in the input validation logic adds significant overhead. If
         set, input_units *must* be a valid unit object. Defaults to False.
+    name : string
+        The name of the array. Defaults to None. This attribute does not propagate
+        through operations.
```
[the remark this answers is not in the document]
l-johnston: done.

unyt/array.py
```diff
@@ -1941,6 +1955,9 @@ class unyt_quantity(unyt_array):
         of the registry associated with the unit object.
     dtype : data-type
         The dtype of the array data.
+    name : string
+        The name of the scalar. Defaults to None. This attribute does not propagate
+        through operations.
```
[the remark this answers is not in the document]
l-johnston: done.

unyt/array.py
```diff
@@ -541,7 +547,7 @@ def __new__(
 
         # Attach the units
```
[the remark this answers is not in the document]
l-johnston: done.

PR #131, opened 2020-01-27
fix UnitOperationError exception when the exponent is unyt_array or u…

l-johnston:
…nyt_quantity type

Raising a value to a unyt_array type causes an AttributeError in the __str__() of UnitOperationError since unit2 is a unyt_array and not a Unit.

```
>>> from unyt import m
>>> 10**(1*m)
  File "/mnt/d/Python/unyt/unyt/unyt/array.py", line 1669, in __array_ufunc__
    raise UnitOperationError(ufunc, u0, u1)
unyt.exceptions.UnitOperationError: <exception str() failed>
```

### DIFF
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -63,7 +63,7 @@ from unyt.unit_symbols import cm, m, g, degree
 from unyt.unit_registry import UnitRegistry
 from unyt._on_demand_imports import _astropy, _h5py, _pint, NotAModule
 from unyt._physical_ratios import metallicity_sun, speed_of_light_cm_per_s
-from unyt import dimensions
+from unyt import dimensions, Unit
 
 
 def operate_and_compare(a, b, op, answer):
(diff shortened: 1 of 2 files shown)


unyt/tests/test_unyt_array.py
```diff
@@ -498,6 +498,12 @@ def test_power():
     with pytest.raises(UnitOperationError):
         np.power(cm_arr, unyt_quantity(3, "g"))
 
+    try:
+        np.power(cm_arr, unyt_quantity(3, "g"))
+    except UnitOperationError as err:
+        assert isinstance(err.unit1, Unit)
+        assert isinstance(err.unit2, Unit)
+
```
ngoldbaum: Good catch on the bug!

Assuming my reading of the pytest docs is correct, I'm pretty sure this can be phrased like this:

```
with pytest.raises(UnitOperationError) as excinfo:
    assert type(excinfo.value.unit1) is Unit
    assert type(excinfo.value.unit2) is Unit
```

However I think the way you have it is actually simpler so I'm going to merge it as-is.

Issue #132, opened 2020-01-28
Unyt does not support the units neper and bel for logarithmic quantities.


l-johnston:
### Feature Proposal
The logarithmic quantities level-of-power and level-of-field are commonly used in engineering, defined in IEC 60027-3, and accepted for use with the SI. I propose defining the dimension ``(logarithmic)`` and add the units bel and neper to Unyt. I have implemented this in my [logarithmic_quantities](https://github.com/l-johnston/unyt/tree/logarithmic_quantity) branch. Some examples with this implementation:
```
>>> import numpy as np
>>> from unyt import Np, B, dB, mW, V
>>> B.dimensions
(logarithmic)
>>> B.get_base_equivalent()
Np
>>> p = [1, 100]*mW
>>> p_0 = 1*mW
>>> level_of_power = np.log10(p/p_0)*B
>>> level_of_power
unyt_array([0., 2.], 'B')
>>> level_of_power.to("dB")
unyt_array([ 0., 20.], 'dB')
>>> f = [1, 10]*V
>>> f_0 = 1*V
>>> level_of_field = 20*np.log10(f/f_0)*dB
>>> level_of_field
unyt_array([ 0., 20.], 'dB')
>>> 10**(level_of_field.v/20)*f_0
unyt_array([ 1., 10.], 'V')
>>> (1*B)/(np.log(10)/2) == 1*Np
array(True)
```
Is the Unyt team interested in this feature?

ngoldbaum:
Seems fine to me!

Probably needs some discussion in the docs though.

PR #133, opened 2020-01-28
Add support for logarithmic quantities and the units neper and bel #132

l-johnston:


ngoldbaum:
So I don't have much experience with these logarithmic units. Does it make sense for someone to do:

```
from unyt import dB

10**(300*dB)
```

Because right now that will raise an error, I think.

unyt/_unit_lookup_table.py
```diff
@@ -426,6 +430,8 @@
         ("psi", ("pounds_per_square_inch",)),
         # dimensionless stuff
         ("dimensionless", ("_", "")),
+        ("B", ("bel",)),
```
ngoldbaum: Does `decibel` work as a unit name now or does that need special treatment?

unyt/tests/test_units.py
```diff
@@ -802,3 +802,15 @@ def test_equal_has_same_hash():
...
     assert hash(a) == hash(b)
     assert hash(b) == hash(c)
+
+
+def test_bel_neper():
+    assert Unit("B").dimensions == Unit("Np").dimensions
+    a = 1 * Unit("B") / (np.log(10) / 2)
+    assert_allclose_units(a.to("Np"), 1 * Unit("Np"))
+    a = 2 * Unit("B")
+    b = 20 * Unit("decibel")
```
ngoldbaum: Ah I see it does :)

l-johnston:
As in all other cases, it is improper to raise a value to quantity and so the error is correct. I state in the usage doc to strip the unit using unyt_array.v before the power operation.

docs/usage.rst
```diff
@@ -224,6 +224,29 @@ The one exception to this rule is for trigonometric functions applied to data wi
   >>> np.sin(45.*degree)
   array(0.70710678)
 
+Logarithmic Quantities and Units
+********************************
+
+The logarithmic quantities level-of-power and level-of-field and the units neper and
+bel are supported. In the next example, we represent the power measurements, p, as
+a logarithmic quantity at reference level, p_ref, in the units decibel.
```
ngoldbaum: `p` and `p_ref` need double backticks, right now I think the underscore in p_ref will cause formatting issues

docs/usage.rst
```diff
@@ -224,6 +224,29 @@ The one exception to this rule is for trigonometric functions applied to data wi
...
+  >>> dB.dimensions
+  (logarithmic)
+  >>> p = [1, 100]*mW
+  >>> p_ref = 1*mW
+  >>> level_of_power = 10*np.log10(p/p_ref)*dB
+  >>> level_of_power
+  unyt_array([ 0., 20.], 'dB')
+
+You can convert the logarithmic quantity back to physical units through exponentiation,
+just remember to remove the units using the unyt_array.v property.
```
ngoldbaum: `unyt_array.v` needs either double backticks or, even better, to be a cross-reference to the API docs.

unyt/_physical_ratios.py
```diff
@@ -160,3 +160,7 @@
 pascal_per_atm = 101325.0
 m_per_inch = 0.0254
 m_per_ft = 0.3048
+
+# logarithmic units
+# IEC 60027-3
```
ngoldbaum: Can you elaborate in this comment more what this is referring to? (e.g. with a link)

docs/usage.rst
l-johnston: done.

docs/usage.rst
l-johnston: done.

unyt/_physical_ratios.py
l-johnston: done.

PR #140, opened 2020-02-13
Fix bugs spotted by the yt tests

ngoldbaum:
The addition of `name` to the `__new__` signature broke existing subclasses(like yt's `ImageArray`) that don't have `name` in their `__new__` signature. As a bandaid I've wrapped the places `__new__` gets called in try/except blocks.

I also fixed a logic error that caused custom units to not be loaded correctly from jsons or pickles.

I've added tests that would catch both of these issues so hopefully we won't have future recurrances of these regressions.

It may make sense to run a test job that runs the yt unit tests since it seems to be able to find more bugs thant he unyt tests alone.

ping @l-johnston for the changes to how `name` works - maybe there's a better way than try/except?

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -824,9 +824,13 @@ class unyt_array(np.ndarray):
             if offset:
                 np.subtract(ret, offset, ret)
 
-            new_array = type(self)(
-                ret, new_units, bypass_validation=True, name=self.name
-            )
+            try:
+                new_array = type(self)(
+                    ret, new_units, bypass_validation=True, name=self.name
+                )
+            except TypeError:
+                # subclasses might not take name as a kwarg
+                new_array = type(self)(ret, new_units, bypass_validation=True)
 
             return new_array
         else:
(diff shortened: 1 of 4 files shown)


l-johnston:
I don't see the problem. Here's what I tried:
```
>>> from unyt import unyt_array
>>> class MyArray(unyt_array):
...     def __new__(cls, input_array, units):
...         return super().__new__(cls, input_array, units)
...
>>> x = MyArray([0,1,2], "V")
>>> x
MyArray([0, 1, 2], 'V')
```
Can you provide a toy example that breaks?

ngoldbaum:
Try calling copy or in_units.

ngoldbaum:
for example:

```
In [3]: from unyt import unyt_array
   ...: class MyArray(unyt_array):
   ...:      def __new__(cls, input_array, units, bypass_validation=False):
   ...:          return super().__new__(cls, input_array, units, bypass_validation=bypass_validation)
   ...:
   ...: x = MyArray([0,1,2], "V")

In [4]: x.to('mV')
TypeError                                 Traceback (most recent call last)
<ipython-input-4-96b7b2bd03c3> in <module>
----> 1 x.to('mV')

~/Documents/unyt/unyt/array.py in to(self, units, equivalence, **kwargs)
    875         898755178736817.6 J
    876         """
--> 877         return self.in_units(units, equivalence=equivalence, **kwargs)
    878
    879     def to_value(self, units=None, equivalence=None, **kwargs):

~/Documents/unyt/unyt/array.py in in_units(self, units, equivalence, **kwargs)
    825                 np.subtract(ret, offset, ret)
    826
--> 827             new_array = type(self)(
    828                 ret, new_units, bypass_validation=True, name=self.name
    829             )

TypeError: __new__() got an unexpected keyword argument 'name'
```

ngoldbaum:
OK, this should be good to merge now.

PR #149, opened 2020-04-13
Including data files in the package directory

migueldvb:
The data file in the tests is not currently not installed in the package directory. Although pytest works fine from the `tests` directory, running the `unyt.test()` function gives this error message:

```
_________________________ test_old_registry_json __________________________

    def test_old_registry_json():
        path = os.sep.join([os.path.dirname(__file__), "old_json_registry.txt"])
>       with open(path) as f:
E       FileNotFoundError: [Errno 2] No such file or directory: '/usr/local/lib/python3.8/site-packages/unyt-2.7.1+2.g57744cc-py3.8.egg/unyt/tests/old_json_registry.txt'

/usr/local/lib/python3.8/site-packages/unyt-2.7.1+2.g57744cc-py3.8.egg/unyt/tests/test_unit_registry.py:97: FileNotFoundError
```

Adding the `package_data` argument to the setup function to specify the files that are included in the package directory fixes the `FileNotFoundError`.


ngoldbaum:
Fair enough, thanks for the fix! If you’re up for it, modifying the tox config to add a test run on only one python version (no need to run it over and over for all versions) that imports unyt and runs unyt.test(), that would  be awesome. I’ll merge this without the test change but it would be nice to have CI for unyt.test() so we don’t regress for another reason in the future.

PR #150, opened 2020-04-14
Adding optional line in tox for python 3.8

migueldvb:
@ngoldbaum I added a conditional line in the tox configuration to run the unyt.test() function for python3.8 (https://tox.readthedocs.io/en/latest/config.html#factors). It seems to work fine on my local environment.

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -1,9 +1,9 @@
 [tox]
-envlist = py36-docs,begin,py35-dependencies,py35-versions,py{35,36,37,38},end
+envlist = py36-docs,begin,py35-dependencies,py35-versions,py{35,36,37,38},py38-unyt-module-test-function,end
 
 [travis]
 python =
-    3.8: py38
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Sorry to nitpick, but can you make this a separate test job, e.g. `py38-unyt-module-test-function`? I think having it only on this one line in the general build command will make it less clear that the failure ultimately comes from calling unyt.test(). If it’s its own test job it will be much more clear from the tox output and the name of the job that failed where the failure comes from.

migueldvb:
got you! I have added a separate test environment for the `unyt.test` function
on python 3.8 that depends on the py38 environment so that they will share the
same dependencies instead of duplicating the same information under the two
tests.  It works when both environments are run together or separately. Does
it look good like this?

ngoldbaum:
Looks good! Thank you for working on this :)

PR #151, opened 2020-05-03
Add special case when printing degree Celsius and fix spelling mistake

l-johnston:


unyt/_unit_lookup_table.py
```diff
@@ -414,7 +414,7 @@
         ("Wb", ("weber",)),
         ("lm", ("lumen",)),
         ("lx", ("lux",)),
-        ("degC", ("degree_celsius", "celcius")),
+        ("degC", ("degree_celsius", "celsius")),
```
ngoldbaum, line 417: I think we need to leave the spelling error in just in case someone was using it. I think we also need to add the spelling with the unicode degree symbol so copy/pasting that in will work (even if it’s hard to type).

It would also be nice to have a sentence or two about these subtleties somewhere in the docs.

ngoldbaum: (and also add the unicode spelling for fahrenheit)

l-johnston: I will add back 'celcius' and then add 'celsius'.

I tried adding '°C' to the tuple of alternative names, but this causes the SyntaxError exception in sympy.
```
>>> from unyt import Unit
>>> Unit("°C")
SyntaxError                               Traceback (most recent call last)
/mnt/d/Python/unyt/unyt/_parsing.py in parse_unyt_expr(unit_expr)
     90     try:
---> 91         unit_expr = parse_expr(
     92             unit_expr, global_dict=global_dict, transformations=unit_text_transform
SyntaxError: invalid character in identifier (<string>, line 1)
```

I am willing to debug further if you'd like.

ngoldbaum: Ah I guess unit names need to be valid python identifiers then. In that case I think you’ll need to add a special case to the parser so it gets canonicalized to `degC`. We do a similar thing for attosecond since it shadows the python `as` keyword.

unyt/unit_object.py
```diff
@@ -331,6 +331,10 @@ def __repr__(self):
     def __str__(self):
         if self.expr == sympy_one:
             return "dimensionless"
+        if self.expr.__str__() == "degC":
```
ngoldbaum, line 334: Sorry that I missed this earlier, can you inline the call to `expr.__str__()`? Printing and string generation can unfortunately sometimes be in performance-sensitive code paths so if we can avoid calling that function three times we should.

PR #152, opened 2020-05-04
Add temperature units delta_degC and delta_degF

### LINKED ISSUE #99: add two temperature values in celcius and Kelvin
* unyt version:  2.2.2
* Python version:   3.7.1
* Operating System: Windows

I just add two temperature values in celcius and Kelvin.
But the result in celcius is wrong

In [16]: from unyt import celcius, K
In [17]: temp1 = 10*celcius
In [18]: temp2 = 1*K
In [19]: temp1+temp2
Out[19]: unyt_quantity(11, 'degC')

Result expected: unyt_quantity(-262,15, 'degC')

l-johnston:
Closes #99

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -356,6 +356,20 @@ The `degC` symbol has alternative names `degree_Celsius`, `Celsius` and `°C`.
 
 These comments also apply to degree Fahrenheit.
 
+Performing arithmetic with temperature quantities can be ambiguous. To clarify intent,
+Unyt has the convenience units `delta_degC` and `delta_degF`.
+
+  >>> from unyt import degC, delta_degC, V
+  >>> t1 = 23*degC
+  >>> t2 = 1*delta_degC
+  >>> print(t1 + t2)
+  24 °C
+  >>> print(t2 - t1)
+  -22 °C
+  >>> tempco = 10.0*V/delta_degC
+  >>> print(tempco*2*delta_degC)
+  20.0 V
+
 Unit Conversions and Unit Systems
 +++++++++++++++++++++++++++++++++
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -335,8 +335,8 @@ using the :meth:`@accepts <unyt.dimensions.accepts>` and :meth:`@returns <unyt.d
 Temperature Units
 -----------------
 
-The temperature unit degree Celsius has the symbol °C, but since the degree character
-is an invalid Python identifier, Uynt uses the symbol `degC`. Printing a degree Celsius
+The temperature unit degree Celsius has the symbol ``°C``, but since the degree character
+is an invalid Python identifier, :mod:`unyt` uses the symbol ``degC``. Printing a degree Celsius
 quantity will show the correct symbol.
 
   >>> from unyt import degC
@@ -344,7 +344,7 @@ quantity will show the correct symbol.
   >>> print(Ta)
   23 °C
 
-The `degC` symbol has alternative names `degree_Celsius`, `Celsius` and `°C`.
+The ``degC`` symbol has alternative names ``degree_Celsius``, ``Celsius`` and ``°C``.
 
   >>> from unyt import degree_Celsius, unyt_array
   >>> Ta = 23*degree_Celsius
@@ -357,7 +357,7 @@ The `degC` symbol has alternative names `degree_Celsius`, `Celsius` and `°C`.
 These comments also apply to degree Fahrenheit.
 
 Performing arithmetic with temperature quantities can be ambiguous. To clarify intent,
-Unyt has the convenience units `delta_degC` and `delta_degF`.
+:mod:`unyt` has the convenience units ``delta_degC`` and ``delta_degF``.
 
   >>> from unyt import degC, delta_degC, V
   >>> t1 = 23*degC
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -164,11 +164,11 @@ def _multiply_units(unit1, unit2):
     return ret.as_coeff_unit()
 
 
-TEMPERATURE_WARNING = (
-    "Ambiguous operation with heterogeneous temperature units.\n"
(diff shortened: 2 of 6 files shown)


ngoldbaum:
Thanks this looks great! There are subtleties to this so I want to take some time and think carefully about if this is correct or not. The implementation looks good though and I don't see significant possible performance concerns (`_preserve_unit` is behind an LRU cache so we only pay any extra cost once).

I updated the description so that #99 gets closed when this is closed.

ngoldbaum:
It might also be nice to look at the impact of deprecating (by issuing a user-visible deprecation warning) the problematic operations @ksunden pointed out in https://github.com/yt-project/unyt/issues/99#issuecomment-519140215.

l-johnston:
The SI standard allows degree Celsius and kelvin to describe temperature intervals and differences - see section 8.5 of [NIST Special Publication 811](https://www.nist.gov/pml/special-publication-811). In the case of binary operations with homogeneous units the result is numerically correct, consistent with the standard and shouldn't cause confusion. The case of heterogeneous units is ambiguous. In this case, Mathcad considers kelvin as the temperature difference since it has an offset of 0. I recommend Unyt do the same and issue a FutureWarning that kelvin will be considered the temperature difference and recommend using delta_degC to clarify intent. In the present implementation this looks like:

```
>>> from unyt import K, degC, delta_degC
>>> t1 = 10*degC
>>> t2 = 1*K
>>> t1 + t2
unyt_quantity(11, 'degC')
>>> t2 + t1
/mnt/d/Python/unyt/unyt/array.py:1765: FutureWarning:
After release x, kelvin and Rankine will be temperature deltas in operations with
heterogenous units. We recommend using delta_degC and delta_degF to clarify intent.

unyt_quantity(11, 'K')
>>> td1 = 1*delta_degC
>>> td1 + t1
unyt_quantity(11, 'degC')
```

l-johnston:
I added the Future Warning. Is this sufficient or do you recommend a different approach?

ngoldbaum:
Sorry I haven’t had time to look at this closely yet, it may be a while. I’m trying to recover from burnout and it takes a lot of mental effort to write or read code right now.

l-johnston:
Ok, focus on recovery.

docs/usage.rst
```diff
@@ -356,6 +356,20 @@ The `degC` symbol has alternative names `degree_Celsius`, `Celsius` and `°C`.
 
 These comments also apply to degree Fahrenheit.
 
+Performing arithmetic with temperature quantities can be ambiguous. To clarify intent,
+Unyt has the convenience units `delta_degC` and `delta_degF`.
```
jzuhone, line 360: I would render this as ``unyt``

jzuhone:
Other than the minor docs nit, I'm good with this.

docs/usage.rst
l-johnston: I updated the formatting.

unyt/array.py
```diff
@@ -164,7 +164,21 @@ def _multiply_units(unit1, unit2):
     return ret.as_coeff_unit()
 
 
+TEMPERATURE_WARNING = """
+After release x, kelvin and Rankine will be temperature deltas in operations with
+heterogeneous units. We recommend using delta_degC and delta_degF to clarify intent.
+"""
```
ngoldbaum, line 170: I think it would be better to just error out in the future when we make the behavior change instead of silently changing results. I think it would be better if we encourage our users to write non-ambiguous expressions instead of allowing them in a different way when we do make this behavior change.

Also instead of saying that we'll change behavior in a specific future release I think it would be safer to just leave the exact version when the behavior change will happen ambiguous (e.g. instead of `After release x` I'd say `In the future`). That way we don't confuse anyone unnecessarily if we forget to actually do the behavior change.

unyt/tests/test_unyt_array.py
```diff
@@ -2379,3 +2379,29 @@ def test_neper_bel():
...
     with pytest.raises(InvalidUnitOperation):
         np.power(unyt_array([0, 20], "dB"), -2)
+
+
+def test_delta_degC():
+    t1 = 10 * degC
+    t2 = 1 * K
+    assert t1 + t2 == 11 * degC
+    with pytest.warns(FutureWarning):
+        t2 + t1
```
ngoldbaum, line 2389: Can we test that the result we get is correct?

unyt/tests/test_unyt_array.py
```diff
@@ -2379,3 +2379,29 @@ def test_neper_bel():
...
+    assert 1 * delta_degC + 2 * delta_degC == 3 * delta_degC
+    assert 2 * delta_degC == unyt_quantity(2, "delta_degC")
+
+
+def test_delta_degF():
+    t1 = 10 * degF
+    t2 = 1 * R
+    assert t1 + t2 == 11 * degF
+    with pytest.warns(FutureWarning):
+        t2 + t1
```
ngoldbaum, line 2402: Same here, we should check that the result is still what we expect.

unyt/array.py
l-johnston: Addressed both concerns and the behavior will appear as:

```
>>> from unyt import K, degC, delta_degC
>>> 10*K + 23*degC
.../array.py:1766: FutureWarning: Ambiguous operation with heterogeneous temperature units.
In the future, such operations will generate UnitOperationError.
Use delta_degC or delta_degF to avoid the ambiguity.
  mul, unit = unit_operator(u0, u1)
unyt_quantity(33, 'K')
>>> 10*delta_degC + 23*degC
unyt_quantity(33, 'degC')
```

unyt/tests/test_unyt_array.py
l-johnston: Added assertion

unyt/tests/test_unyt_array.py
l-johnston: Added assertion

PR #155, opened 2020-06-29
add more alternative spellings for solar units

ngoldbaum:
Fixes https://github.com/yt-project/yt/issues/2704.

Added a test too.


neutrinoceros:
great !
While we're at it, I was also considering adding `"m_sun"` as well as `"tsun"` for consistency with the different solar units. What do you think ?

ngoldbaum:
Can you check that those don’t already work? The logic that generates names is more complicated than what’s in these lists.

neutrinoceros:
Just did, I confirm that none of them currently work.

ngoldbaum:
OK, I've gone ahead and added a test that systematically tests for various alternative spellings for solar units. I'd appreciate it if @jzuhone could sanity check what I've added before this gets merged. I'll go ahead and do the unyt release afterwards.

munkm:
Thanks @ngoldbaum !

PR #156, opened 2020-06-30
raise TypeError on bad input instead of RuntimeError in unyt_quantity init

neutrinoceros:
A simple proposal to raise a somewhat more specific error upon bad input data when initialising a `unyt_quantity` object
I was writing some tests for yt to check that we caught this kind of typing error before passing them down to unyt.
Admittedly this is a breaking change for downstream code relying on RuntimeError to be raised, though I hope this is minor enough that it can be accepted. If there's a smoother path to perform such a transition ? Anything I can do to make this better ? Let me know !

for reference, the yt tests in questions are there https://github.com/yt-project/yt/pull/2703

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -2022,22 +2022,24 @@ class unyt_quantity(unyt_array):
             bypass_validation
             or isinstance(input_scalar, (numeric_type, np.number, np.ndarray))
         ):
-            raise RuntimeError("unyt_quantity values must be numeric")
+            raise TypeError("unyt_quantity values must be numeric")
         if input_units is None:
             units = getattr(input_scalar, "units", None)
         else:
             units = input_units
+        scalar_array = np.asarray(input_scalar)
+        if scalar_array.size != 1:
+            raise TypeError("input_scalar must be a single scalar")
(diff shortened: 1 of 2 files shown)


ngoldbaum:
I don’t think this is worth the backward incompatibility break. What if someone has code that catches `RuntimeError`? This change would break that code.

neutrinoceros:
Yes it's a possibility. One way to transition between release would be to raise a custom error that derives from both RuntimeError and TypeError, though it certainly adds maintenance overhead.

ngoldbaum:
You could make a new exception type that inherits from both `RuntimeError` and `TypeError` but I don’t think that’s worth the code churn either.

tbh I’m inclined to close this PR unless someone can demonstrate to me that this is worth the code churn.

neutrinoceros:
I'm going to stand back for now. I may have rushed through this and I'm having a hard time arguing in favour of the backwards compatibility break. I still think this is niche enough that I don't anticipate it to be a major disruption, but I admit I can not measure or demonstrate this. You may close this if you wish, and I won't reopen without a solid argument 😊

PR #158, opened 2020-07-06
Replace dimensions that are equal but not identical.

### LINKED ISSUE #157: Crash with LaTeX representation and multiple unit registry loads
* unyt version: 2.7.2 (released)
* Python version: 3.7
* Operating System: Linux

### Description

This showed up during the yt test suite, but I was *mostly* able to reduce it to a self-contained error.

It seems that after some combination of events, which can be triggered by calling `latex_representation()` on a combination of two `Unit` objects, can cause state to be retained that doesn't allow for subsequent unit registries to be created.

### What I Did

This shows up if you load the ytdata dataset `slice.h5` after doing *anything* that calls `latex_representation` on another dataset (i.e., making a plot), but I've reduced it here, along with an additional file that is the json unit registry from the ytdata file.  I've created a gist with the JSON information here: https://gist.github.com/matthewturk/4fdf1335c8aa5df315f838d47c625bc7 .

```
import unyt

reg1 = unyt.UnitRegistry()
reg1.add("code_length", 1.0, unyt.dimensions.length)
reg1.add("code_mass", 1.0, unyt.dimensions.mass)
reg1.add("code_time", 1.0, unyt.dimensions.time)
reg1.add("code_temperature", 1.0, unyt.dimensions.temperature)
us1 = unyt.UnitSystem(reg1.unit_system_id, "code_length", "code_mass", "code_time",
                     "code_temperature", registry = reg1)

cm = unyt.Unit('code_mass', registry=reg1)
cl = unyt.Unit('code_length', registry=reg1)

(cm / cl).latex_representation()

reg2 = unyt.UnitRegistry.from_json(open("pp_ureg.json").read())
us2 = unyt.UnitSystem(reg2.unit_system_id, "code_length", "code_mass", "code_time",
                     "code_temperature", registry = reg2)

```

This produces:

```
  File "breakit.py", line 20, in <module>
    "code_temperature", registry = reg2)
  File "/home/mturk/yt/unyt/unyt/unit_systems.py", line 233, in __init__
    raise IllDefinedUnitSystem(self.units_map)
unyt.exceptions.IllDefinedUnitSystem: Cannot create unit system with inconsistent mapping from dimensions to units. Received:
OrderedDict([((length), code_length), ((mass), code_mass), ((time), code_time), ((temperature), code_temperature), ((angle), rad), ((current_mks), A), ((luminous_intensity), cd), ((logarithmic), Np)])
```

@xarthisius and I tried pretty hard to track this down, but was unable to fully figure it out; for instance, in the `LatexPrinter`, using `str(l_expr)` instead didn't trigger it, etc etc.  But what I *suspect* and am not entirely sure of, is that there's an `is` comparison that is not behaving as expected as a result of `_correct_old_unit_registry`.  On line 230 or so of `unit_systems.py` there's this:

```
            if (
                self.registry is not None
                and self.registry[str(unit)][1] is not dimension
            ):
```

but it looks like:

```
>>> id(reg2['code_length'][1])
139744493368576
>>> id(reg1['code_length'][1])
139744510534432
```

So an `==` check passes but `is` does not; I assume this is behaving as expected, and that somehow we're creating two different `length` dimensions, somewhere.

Anyway!  I see a couple possible solutions.  One would be to, in `_correct_old_unit_registry`, replace all the `unsan_v[1]` values that are *equal* to contents of `unyt_dims` with those `unyt_dims` values.  (I checked, and the `==` passes but `is` does not, so this seems to be a pain point.)  The other might be to replace the `is` comparison in `unit_systems.py` with `==`.

matthewturk:
This closes #157 by examining the old-style dimensions, and if they are *equal* but not *identical*, replacing them.  This allows them to match the dimensions checked in the unit registry later.

### DIFF
--- a/unyt/tests/test_unit_registry.py
+++ b/unyt/tests/test_unit_registry.py
@@ -19,10 +19,11 @@ import pytest
 
 from numpy.testing import assert_allclose
 
-from unyt.dimensions import length, energy
+from unyt.dimensions import length, energy, mass, time, temperature
 from unyt.exceptions import SymbolNotFoundError, UnitParseError
 from unyt.unit_object import Unit
 from unyt.unit_registry import UnitRegistry
+from unyt.unit_systems import UnitSystem
 
 
 def test_add_modify_error():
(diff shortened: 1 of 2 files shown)


matthewturk:
Whoops, definitely not ready.  Talk about a misread of the code!

matthewturk:
I *think* this version should work.  The tests pass, and the yt tests pass as well, but alternate suggestions are appreciated.

ngoldbaum:
Sorry for taking a couple days to respond. I looked at this and spotted some issues but wanted to understand the issue and code a little more clearly myself so I decided to rewrite this a little bit to make it clearer for me and not respond until I had time to do that. 

I added a test that fails on master using the existing old JSON unit registry already used by the tests so hopefully we won't regress on this.

I'd appreciate it if you or @Xarthisius or whoever else could verify that the change as I have it here does fix the yt issue in the same way as your original version of this pull request.

I didn't have time to run the tests locally so they might fail, if so I'll fix that ASAP and ping you. Unfortunately I need to accompany Shawna to the prosthetist this afternoon so I won't be able to do that for a few hours.

Hope you all are having a good week and enjoyed scipy :)

Xarthisius:
I checked-out `+refs/pull/158/merge:` and both yt test case and unyt #157 work for me now. Thanks @ngoldbaum !

PR #160, opened 2020-07-10
Add the imperial unit mil.

l-johnston:
The mil, 0.001 inch, is in common use here in the U.S. for printed circuit board design and production.


ksunden:
Sometimes this is also referred to as the "thou", to avoid confusion with mm

And in machining, a "tenth" refers, confusingly, to a "tenth of a thou" i.e. 0.0001 inch

ngoldbaum:
How about adding `thou` and `thousandth` as alternate spellings for this via an entry in `default_unit_name_alternatives`. It looks like a US vs imperial thing to call this `thou` or `mil`.

PR #162, opened 2020-07-21
Add US customary units kip and ksi

otaithleigh:
Adds the force unit kip (a.k.a. kipf or kilopound, 1000 lbf) and the pressure unit ksi (kips per square inch, 1000 psi). These are most commonly used in architecture and engineering. I'm following `pint` in making the default name kip refer to a force unit.

I'm a little confused on tests -- some are in `test_units.py` and some are in `test_unyt_array.py`. Is there a preferred place for tests for units?


ngoldbaum:
`test_units.py` is probably best

otaithleigh:
Updated with tests. I put them in `test_unyt_array.py` since I realized that `test_units.py` is being used for the `str` representation.

PR #163, opened 2020-07-31
bugfix: fix commutativity in unyt_array operators

neutrinoceros:
fix https://github.com/yt-project/yt/issues/874

Here's an actualised version of the script provided by @jzuhone at the time with updated reference outputs.
```python

>>> import yt
>>> ds = yt.testing.fake_amr_ds()
>>> a = yt.YTArray([1,2,3], "cm")
>>> b = ds.arr([1,2,3], "code_length")

>>> a*b
old > SymbolNotFoundError: The symbol 'code_length' does not exist in this registry.
new > unyt_array([1, 4, 9], 'cm*code_length')

>>> b*a
old > unyt_array([1, 4, 9], 'cm*code_length')
new > unyt_array([1, 4, 9], 'cm*code_length')

>>> (a*b).in_units("code_length**2")
old > SymbolNotFoundError: The symbol 'code_length' does not exist in this registry.
new > unyt_array([1., 4., 9.], 'code_length**2')

>>> (b*a).in_units("code_length**2")
old > unyt_array([1., 4., 9.], 'code_length**2')
new > unyt_array([1., 4., 9.], 'code_length**2')
```

For reference, this issue was referenced in https://github.com/yt-project/yt/issues/2797, hence the fix.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -123,6 +123,7 @@ from unyt.exceptions import (
     UnitOperationError,
     UnitConversionError,
     UnitsNotReducible,
+    SymbolNotFoundError,
 )
 from unyt.equivalencies import equivalence_registry
 from unyt._on_demand_imports import _astropy, _pint
@@ -1763,7 +1764,13 @@ class unyt_array(np.ndarray):
                         )
                     inp1 = np.asarray(inp1, dtype=new_dtype) * conv
             # get the unit of the result
-            mul, unit = unit_operator(u0, u1)
+            try:
+                mul, unit = unit_operator(u0, u1)
+            except SymbolNotFoundError:
+                # Some operators are not natively commutative when operands are
+                # defined within different unit registries, and conversion
+                # is defined one way but not the other.
+                mul, unit = unit_operator(u1, u0)
             # actually evaluate the ufunc
             out_arr = func(
                 inp0.view(np.ndarray), inp1.view(np.ndarray), out=out_func, **kwargs


ngoldbaum:
Won’t this break with a non-commutative operator like division?

neutrinoceros:
Oh you're right...
So I guess the correct way to do it would be to fix each commutative operator separately, or maybe add a decorator for this to avoid code duplication ?
Unless non-commutative operators are the exception rather than the rule, in which case maybe they could be special-cased ?

neutrinoceros:
another (better ?) approach would be to try conversions uphill (as a sanitizing step) and only then pass the objects to the operator ?

ngoldbaum:
Please send a followon if you feel like it. Gonna close this since this approach is a dead-end.

neutrinoceros:
I think I have a better fix. Is it okay to reopen this PR (with a rewritten history) or should I open a new one ?

ngoldbaum:
Please open a new one.

ngoldbaum:
Also please include tests.

neutrinoceros:
> Please open a new one.

sure thing

> Also please include tests.

I'd like to but I don't know how to reproduce this without yt. My knowledge of unyt's and in particular unit registries only goes so far...

ngoldbaum:
Please read the docs, they clearly explain how to define custom units on a unit registry (as yt is doing internally). To be honest if you aren’t willing to put in the work on this to get your contribution merged then this process isn’t going to go well.

ngoldbaum:
There are plenty of examples of this in the unyt tests too.

neutrinoceros:
> To be honest if you aren’t willing to put in the work on this to get your contribution merged then this process isn’t going to go well.

Please give me a minute Nathan... I'm trying to help, I'd appreciate if you would please try to be just a little more supportive in your answers. I want to do this but this kind of message is really killing my will to go through.

ngoldbaum:
It kills my will to help when I see that the initial approach is obviously incorrect and doesn’t include tests.

jzuhone:
@neutrinoceros go ahead and open another PR with your new approach and I can take a look. I do agree with @ngoldbaum that we need some tests though.

PR #164, opened 2020-07-31
bugfix: fix commutativity in unyt_array operators

neutrinoceros:
another go at https://github.com/yt-project/unyt/pull/163, and fix https://github.com/yt-project/yt/issues/874

I've added tests for both commutative and non-commutative operators. Two of them (`//` and `%`) are currently failing and I can't seem to understand why, could it be due to rounding errors ?
@jzuhone , do you have any clue ?

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -123,6 +123,7 @@ from unyt.exceptions import (
     UnitOperationError,
     UnitConversionError,
     UnitsNotReducible,
+    SymbolNotFoundError,
 )
 from unyt.equivalencies import equivalence_registry
 from unyt._on_demand_imports import _astropy, _pint
@@ -160,7 +161,13 @@ def _sqrt_unit(unit):
 
 @lru_cache(maxsize=128, typed=False)
 def _multiply_units(unit1, unit2):
-    ret = (unit1 * unit2).simplify()
+    try:
+        ret = (unit1 * unit2).simplify()
+    except SymbolNotFoundError:
+        # Some operators are not natively commutative when operands are
+        # defined within different unit registries, and conversion
+        # is defined one way but not the other.
+        ret = (unit2 * unit1).simplify()
     return ret.as_coeff_unit()
 
 
@@ -195,7 +202,10 @@ def _square_unit(unit):
 
 @lru_cache(maxsize=128, typed=False)
 def _divide_units(unit1, unit2):
-    ret = (unit1 / unit2).simplify()
+    try:
+        ret = (unit1 / unit2).simplify()
+    except SymbolNotFoundError:
+        ret = (1/(unit2 / unit1).simplify()).units
     return ret.as_coeff_unit()
 
 
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -205,7 +205,7 @@ def _divide_units(unit1, unit2):
     try:
         ret = (unit1 / unit2).simplify()
     except SymbolNotFoundError:
-        ret = (1/(unit2 / unit1).simplify()).units
+        ret = (1 / (unit2 / unit1).simplify()).units
     return ret.as_coeff_unit()
 
 
--- a/unyt/tests/test_units.py
+++ b/unyt/tests/test_units.py
@@ -885,11 +885,7 @@ def test_mixed_registry_operations():
     b = unyt_quantity(1, "cm")
 
     assert a + b == b + a
-    assert a - b == - (b - a)
+    assert a - b == -(b - a)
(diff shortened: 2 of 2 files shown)


jzuhone:
isn't `//` integer division? It's not clear to me if `//` even makes sense in this context. But can you elaborate on the failure?

neutrinoceros:
> isn't // integer division?

it is !

> It's not clear to me if // even makes sense in this context. But can you elaborate on the failure?

sure. So right now this is what the test is getting, and comparing
```python
>>> print(b//a, b//a.in_units("cm"))
10.0 dimensionless 9.0 dimensionless
```

The fact that `b//a.in_units("cm")` is already a valid statement on master is the reason for this test.

neutrinoceros:
Actually this might come from a different bug with the `//` operator, see this script (as ran on master)

```python
import unyt
a = unyt.unyt_quantity(0.1, "km")
b = unyt.unyt_quantity(1, "km")
for unit in ("cm", "m", "km"):
    print(b%a.in_units(unit), b//a.in_units(unit))
```

output
```
0.09999999999999995 km 0.0 dimensionless
0.09999999999999995 km 0.0 dimensionless
0.09999999999999995 km 9.0 dimensionless
```

neutrinoceros:
Probably not a unyt bug since
```python
print(0.01//0.001)
print(1.//0.1)
```
output
```
10.0
9.0
```
So I think it's safe to conclude that what is happening here is out of `unyt`'s control and has to do with floating point arithmetics. I'll try to write a more sensible test.

neutrinoceros:
Welp, after many attempts I've come to think that floor div and int div behave badly with unyt_quantity objects, and fixing those is way out of scope, so for now I'm just dropping the tests for those, I hope this is okay...
Please let me know if there's anything more I can do here.

jzuhone:
Can you do some speed tests as @ngoldbaum mentioned?

unyt/array.py
```diff
@@ -195,7 +202,10 @@ def _square_unit(unit):
 
 @lru_cache(maxsize=128, typed=False)
 def _divide_units(unit1, unit2):
-    ret = (unit1 / unit2).simplify()
+    try:
+        ret = (unit1 / unit2).simplify()
+    except SymbolNotFoundError:
+        ret = (1 / (unit2 / unit1).simplify()).units
```
jzuhone, line 208: I think there's a dimensionless `1` unit we could use here? Not sure if that would be better or not.

neutrinoceros: I can try. How would that work ?

neutrinoceros:
so here's an attempt at speed tests using IPython + Jupyter cell magic `%%timeit`

```python
# setup
import numpy as np
import unyt as u

a0 = np.random.normal(0, 1, int(1e8)).reshape(100,100,100,100)
b0 = np.random.normal(0, 1, int(1e8)).reshape(100,100,100,100)
reg = u.UnitRegistry(unit_system="cgs")
reg.add("fake_length", 0.001, u.dimensions.length)
a = u.unyt_array(a0, units="fake_length", registry=reg)
b = u.unyt_array(b0, "cm")


# tests
%%timeit
a/b
master      > 837 ms ± 23.6 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
this branch > 810 ms ± 103 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)

%%timeit
b/a
master      > N/A
this branch > 764 ms ± 6.55 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


%%timeit
a*b
master      > 407 ms ± 13.8 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
this branch > 397 ms ± 13.1 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)


%%timeit
b*a
master      > N/A
this branch > 422 ms ± 28 ms per loop (mean ± std. dev. of 7 runs, 1 loop each)
```

This looks to me like there's no _obvious_ regression, though the accuracy seems pretty bad, and I don't know how to improve it :/

unyt/array.py
ngoldbaum: This is fine.

ngoldbaum:
The changes are behind a LRU cache so performance checks aren’t necessary.

By the way there are benchmarks in the repo that do a better job of controlling for noise than timeit, I wouldn’t ever use timeit for benchmarking anything real.

ngoldbaum:
Clement, in the future if you don’t want to put in the time to work on something in this repo please just open an issue.

Issue #165, opened 2020-08-05
Graceful fallback for unit errors (warning, casting to float/array, etc.) rather than raising UnitOperationError?


mkhorton:
### Description

This is a question/feature request, not a bug. If there is a more appropriate place to ask this, please let me know.

We're looking at a way to add comprehensive unit integration into an existing project. This project is already seeing existing use by other downstream code, so we would prefer not to break backwards compatibility abruptly for our users, and would prefer to support a grace period if possible.

Currently, to take an example:

```
from unyt import Angstrom
a = 3*Angstrom
a + 1  # raises `UnitOperationError`
```

Is it possible to configure `unyt` to instead issue a warning here instead of raising and to return a de-unitized value, e.g. for this example to return `4` as a float?

If this is not possible, is there a single place in the `unyt` code where such a feature could be added?

If this was possible, we would be able to switch to using `unyt` in our project, but disable strict enforcement of units by default for our users so that we wouldn't break any existing scripts overnight. It'd also allow us to have a more gradual transition period, while simultaneously allowing us to strictly enforce units in our testing and CI and when developing new features.

mkhorton:
Naively, looking at the code, I'd think something like replacing:

```
UnitOperationError(ufunc, u0, u1)
```

with:

```
def raise_or_warn(ufunc, u0, u1):
   # ... if raise ...
       raise UnitOperationError(...)
   # ... if warn ...
       warn(...)
       return ufunc(u0, u1)
```

might be sufficient, but I'm not sure if this would be missing important consequences or edge-cases.

ngoldbaum:
My big concern about this is that if library A and B both depend on unyt, then library A’s choice to turn unit errors into warnings shouldn’t affect library B’s choice. In other words I’m somewhat dubious about a single global flag for controlling this behavior.

One way to do this that could work is to make this controllable at the level of the unit registry object. Your library could then create its own set of units that use a custom unit registry that then controls whether or not an error gets raised.

ngoldbaum:
See this section of the unyt docs for more drtails about wrapping unyt: https://unyt.readthedocs.io/en/stable/usage.html#integrating-unyt-into-a-python-library

ngoldbaum:
By the way, any new changes need tests and new functionality need docs, I likely won’t review pull requests implementing this unless there are docs and tests.

mkhorton:
[...]

PR #166, opened 2020-08-05
[WIP] Addition of `UnitOperationWarning` and `strict` kwarg to `UnitRegistry`

mkhorton:
As discussed in https://github.com/yt-project/unyt/issues/165.

Currently have a demo working, however it's failing `test_ufunc` specifically, because it seems like the `u0` or `u1` in this case don't have a `.registry` or `.dimensions` attributes despite being a `unyt_array`. Suggestions here welcome, I'm not sure why this is the case.

Once it's working I can add some additional tests and check style.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -124,6 +124,7 @@ from unyt.exceptions import (
     UnitConversionError,
     UnitsNotReducible,
     SymbolNotFoundError,
+    UnitOperationWarning
 )
 from unyt.equivalencies import equivalence_registry
 from unyt._on_demand_imports import _astropy, _pint
@@ -154,16 +155,18 @@ def _iterable(obj):
         return False
     return True
 
+
 def _unit_operation_error_raise_or_warn(ufunc, u0, u1, func, *inputs):
-    if STRICT:  # True by default
+    if u0.registry.strict and u1.registry.strict:  # True by default
         raise UnitOperationError(ufunc, u0, u1)
     else:
-        warnings.warn("Performing operation without units!")
+        warnings.warn(UnitOperationWarning(ufunc, u0, u1))
         unwrapped_inputs = [
-            i.value if isinstance(i, unyt_array) or isinstance(i, unyt_quantity)
+            i.value if isinstance(i, unyt_array)
             else i for i in inputs
         ]
         return func(*unwrapped_inputs)
+
 
 @lru_cache(maxsize=128, typed=False)
 def _sqrt_unit(unit):
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -139,6 +139,7 @@ from unyt.unit_registry import (
 
 NULL_UNIT = Unit()
 POWER_SIGN_MAPPING = {multiply: 1, divide: -1}
+STRICT = True
 
 __doctest_requires__ = {
     ("unyt_array.from_pint", "unyt_array.to_pint"): ["pint"],
@@ -153,6 +154,16 @@ def _iterable(obj):
         return False
     return True
 
+def _unit_operation_error_raise_or_warn(ufunc, u0, u1, func, *inputs):
+    if STRICT:  # True by default
+        raise UnitOperationError(ufunc, u0, u1)
+    else:
+        warnings.warn("Performing operation without units!")
+        unwrapped_inputs = [
+            i.value if isinstance(i, unyt_array) or isinstance(i, unyt_quantity)
+            else i for i in inputs
+        ]
+        return func(*unwrapped_inputs)
 
 @lru_cache(maxsize=128, typed=False)
 def _sqrt_unit(unit):
@@ -1704,12 +1715,12 @@ class unyt_array(np.ndarray):
             elif ufunc is power:
                 u1 = inp1
                 if inp0.shape != () and inp1.shape != ():
-                    raise UnitOperationError(ufunc, u0, u1)
+                    return _unit_operation_error_raise_or_warn(ufunc, u0, u1, func, *inputs)
                 if isinstance(u1, unyt_array):
                     if u1.units.is_dimensionless:
                         pass
                     else:
-                        raise UnitOperationError(ufunc, u0, u1.units)
+                        return _unit_operation_error_raise_or_warn(ufunc, u0, u1.units, func, *inputs)
                 if u1.shape == ():
                     u1 = float(u1)
                 else:
@@ -1759,9 +1770,9 @@ class unyt_array(np.ndarray):
                                         ret = bool(ret)
                                     return ret
                                 else:
-                                    raise UnitOperationError(ufunc, u0, u1)
+                                    return _unit_operation_error_raise_or_warn(ufunc, u0, u1, func, *inputs)
                         else:
-                            raise UnitOperationError(ufunc, u0, u1)
+                            return _unit_operation_error_raise_or_warn(ufunc, u0, u1, func, *inputs)
                     conv, offset = u1.get_conversion_factor(u0, inp1.dtype)
                     new_dtype = np.dtype("f" + str(inp1.dtype.itemsize))
                     conv = new_dtype.type(conv)
--- a/unyt/exceptions.py
+++ b/unyt/exceptions.py
@@ -47,6 +47,29 @@ class UnitOperationError(ValueError):
         return err
 
 
+class UnitOperationWarning(UserWarning):
+    """A warning that is raised when unit operations are not allowed but
+    when running with a UnitRegistry with strict=False. In this case,
+    the operation is allowed to continue but with unit information stripped.
+    """
+
+    def __init__(self, operation, unit1, unit2=None):
+        self.operation = operation
+        self.unit1 = unit1
+        self.unit2 = unit2
+        UserWarning.__init__(self)
+
+    def __str__(self):
+        err = (
+            'The %s operator for unyt_arrays with units "%s" '
+            '(dimensions "%s") ' % (self.operation, self.unit1, self.unit1.dimensions)
+        )
+        if self.unit2 is not None:
+            err += 'and "%s" (dimensions "%s") ' % (self.unit2, self.unit2.dimensions)
+        err += "is not well defined. Performing operation without units instead."
+        return err
+
+
 class UnitConversionError(Exception):
     """An error raised when converting to a unit with different dimensions.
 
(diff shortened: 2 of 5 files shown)


mkhorton:
The second aspect of this that would be nice to address:

As written, this PR will always raise a warning. Really, we only want the warning if units on one side of the operator are not specified and assumed dimensionless. If units are actually provided on both sides of the operator, it seems sensible to still raise the `UnitOperationError` and not the warning.

Unfortunately I can't see a way to figure out if the units are intended (i.e. a unyt_array that was intentionally created as dimensionless) or the units are dimensionless simply because they weren't specified and were assumed, i.e. as in:

https://github.com/yt-project/unyt/blob/14e4cdbeeaf6a27ff8a678a2f0a23d32bf8f1b64/unyt/array.py#L562

mkhorton:
Would a `unyt` maintainer be willing to help me understand why `test_ufunc` is failing in this PR? Otherwise the functionality in this PR is fairly complete.

ngoldbaum:
Hey I’ll try to take a look at this soon. Sorry for the radio silence, it’s hard for me to focus on programming tasks these days...

mkhorton:
Many thanks @ngoldbaum, and no worries at all. Grateful for any feedback

neutrinoceros:
Hey @mkhorton, I can help you with this. Since there is a conflict and unyt's CI was migrated to a different system, I'd advise you rebase this branch if you still want to work on this PR. Sorry it's been a while !

mkhorton:
Hi @neutrinoceros, thanks for the offer to help! Would be very happy to see this functionality added.

I've resolved the current merge conflict and this branch is now up to date.

neutrinoceros:
Alright, seems like there are 3 failures, 2 of which are code-style-related. Can you run black + flake8 and fix the warnings you get from the latter ? Then all that should be left is the ufunc issue.

neutrinoceros:
Sorry the infrastructure isn't very well specified yet, it looks like you ran a different version of black that's conflicting with the latest. It'd be preferable to rebase and force push your branch now, to avoid tarnishing the history of files affected by cancelling changes. If you're not confortable with the process I can do it for you, just let me know.

mkhorton:
Ok, I rebased, by all means make changes directly to my fork if preferred. I don't mind installing a specific version of black if necessary too.

neutrinoceros:
We can always squash and rebase later to minimise disrupting your workflow. In the mean time I've pushed a single commit to fix the formatting issue. It is getting late in my timezone, I'll have a deeper look at the actual content tomorrow and hopefully we can move forward with this ufunc issue. Thank you so much for your patience.

unyt/array.py
```diff
@@ -125,6 +125,7 @@
     UnitConversionError,
     UnitsNotReducible,
     SymbolNotFoundError,
+    UnitOperationWarning,
```
neutrinoceros, line 128: Though import ordering isn't automatically enforced atm, can you please reorder this one to preserve alphabetical order here ?

unyt/exceptions.py
```diff
@@ -47,6 +47,29 @@ def __str__(self):
...
+        self.unit2 = unit2
+        UserWarning.__init__(self)
+
+    def __str__(self):
+        err = (
+            'The %s operator for unyt_arrays with units "%s" '
+            '(dimensions "%s") ' % (self.operation, self.unit1, self.unit1.dimensions)
+        )
+        if self.unit2 is not None:
+            err += 'and "%s" (dimensions "%s") ' % (self.unit2, self.unit2.dimensions)
```
neutrinoceros, line 68:
```suggestion
            err += f" and {self.units2!r} (dimensions {self.units2.dimenstions!r})"
```

unyt/exceptions.py
```diff
@@ -47,6 +47,29 @@ def __str__(self):
...
+        err += "is not well defined. Performing operation without units instead."
```
neutrinoceros, line 69:
```suggestion
        err += " is not well defined. Performing operation without units instead."
```

unyt/exceptions.py
```diff
@@ -47,6 +47,29 @@ def __str__(self):
...
+    def __init__(self, operation, unit1, unit2=None):
+        self.operation = operation
+        self.unit1 = unit1
+        self.unit2 = unit2
+        UserWarning.__init__(self)
+
+    def __str__(self):
+        err = (
+            'The %s operator for unyt_arrays with units "%s" '
+            '(dimensions "%s") ' % (self.operation, self.unit1, self.unit1.dimensions)
```
neutrinoceros, line 65: in reprs and error messages, `'` is the preferred delimiter for strings, over `"`. `!r` ensures that we use the raw string repr (`'string'`), and we can use fstrings too now !
```suggestion
            f"The {self.operation.__name__} operator for unyt_arrays"
            f" with units {self.unit1!r} (dimensions {self.unit1.dimensions!r})"
```

unyt/tests/test_unyt_array.py
```diff
@@ -57,6 +57,7 @@
     UnitOperationError,
     UnitParseError,
     UnitsNotReducible,
+    UnitOperationWarning,
```
neutrinoceros, line 60: ordering

unyt/array.py
```diff
@@ -140,6 +141,7 @@
 
 NULL_UNIT = Unit()
 POWER_SIGN_MAPPING = {multiply: 1, divide: -1}
+STRICT = True
```
neutrinoceros, line 144: I haven't found a single place where this is used. Looks like a leftover from a previous idea, can you clean it (or justify its presence ?)

unyt/unit_registry.py
```diff
@@ -67,6 +69,12 @@ def __init__(self, add_default_symbols=True, lut=None, unit_system=None):
         if add_default_symbols:
             self.lut.update(default_unit_symbol_lut)
 
+        # This boolean determines whether to raise a UnitOperationError or
+        # strip units and provide a UnitOperationWarning if an invalid
+        # operation is attempted. The default is strict=True and this is
```
neutrinoceros, line 74:
```suggestion
        # operation is attempted. The default is strict=True and is
```

unyt/array.py
```diff
@@ -166,6 +168,15 @@ def _iterable(obj):
     return True
 
 
+def _unit_operation_error_raise_or_warn(ufunc, u0, u1, func, *inputs):
+    if u0.registry.strict and u1.registry.strict:  # True by default
+        raise UnitOperationError(ufunc, u0, u1)
+    else:
+        warnings.warn(UnitOperationWarning(ufunc, u0, u1))
+        unwrapped_inputs = [i.value if isinstance(i, unyt_array) else i for i in inputs]
+        return func(*unwrapped_inputs)
```
neutrinoceros, line 177: So I can see two problems here
- `unyt_array` doesn't have a `registry` attribute. I think you mean `u0.units.registry`
- even applying the naive fix I get a failing test (in a case where u1 is a pure np.ndarray)

I think I got the fix by reversing the conditional and adding a more complete condition (all tests pass on my machine with this patch)
```suggestion
    if (
        hasattr(u0, "units")
        and not u0.units.registry.strict
        and hasattr(u1, "units")
        and not u1.units.registry.strict
    ):
        warnings.warn(UnitOperationWarning(ufunc, u0, u1))
        unwrapped_inputs = [i.value if isinstance(i, unyt_array) else i for i in inputs]
        return func(*unwrapped_inputs)
    else:
        raise UnitOperationError(ufunc, u0, u1)
```
I have further comments I want to make here but for now let's focus on passing tests, then we'll refactor :)

neutrinoceros (changes requested):
Okay I think I found the fix to your problem. I also have a couple suggestions about string formatting an other small nitpicks. I'd like to review this again after my suggestions are dealt with. Thanks again for your patience, really appreciate it.

unyt/tests/test_unyt_array.py
mkhorton: Saw there were a few others out of order here or incorrectly grouped here too, re-grouped to follow https://www.python.org/dev/peps/pep-0008/#imports

unyt/array.py
mkhorton: Indeed, corrected, thank you! I think this was a remnant from when I was trying to better understand the code and trialling different solutions

unyt/array.py
mkhorton: Thank you kindly for all the comments! These should be resolved now, happy to discuss refactor/changes

unyt/array.py
```diff
@@ -166,6 +167,20 @@ def _iterable(obj):
     return True
 
 
+def _unit_operation_error_raise_or_warn(ufunc, u0, u1, func, *inputs):
```
neutrinoceros, line 170: I am hesitant about how this function implements the proposed functionality. Specifically, it feels weird to replace a raise statement with a return statement (where this function is used).
Another comment is that the `func` argument is redundant (I think), since it can be obtained as `func = ufunc.method`.

neutrinoceros (commented):
So I must say I do not disapprove of the idea, but I have mixed feelings about the current implementation. I'm thinking that Python has a builtin mechanism to change warnings into errors (see https://docs.python.org/3/using/cmdline.html#cmdoption-w), so _maybe_ it's worth considering an alternative implementation where we _always_ raise warnings instead of errors for invalid operations, and so developers of projects that consume the unyt library can choose when they want to turn the `-Werror` flag on in their CI (note that pytest, ipython and other CI relevant interpreters have support for `-W` and pass it down to python). In my opinion this alternative is worth considering, but it needs further discussion.
@ngoldbaum @jzuhone @matthewturk please weigh in.

unyt/unit_registry.py
```diff
@@ -55,7 +55,9 @@ class UnitRegistry:
 
     _unit_system_id = None
 
-    def __init__(self, add_default_symbols=True, lut=None, unit_system=None):
+    def __init__(
+        self, add_default_symbols=True, lut=None, unit_system=None, strict=True
```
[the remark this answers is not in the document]
mkhorton: I hadn't seen this pattern before (using `*` to explicitly make a new argument keyword-only), neat.

unyt/array.py
mkhorton: I share this uneasiness but I'm not sure of a better solution. It seems clear to me that some users will want this to fail explicitly and raise, whereas others will just want the warning, so there has to be a conditional somewhere that switches between that behaviour?

mkhorton: To your second comment, makes sense, I'll remove the `func` argument.

neutrinoceros:
@mkhorton Looks like both of your latest commits broke some tests, can you revert back to the last working state ?
(should be as simple as
```bash
git reset --hard HEAD~2
git push --force
```

PR #170, opened 2020-08-22
Fix unyt_array.__getitem__ to support numpy masked arrays

### LINKED ISSUE #169: Creating masked array raises AttributeError
* unyt version: 2.7.2
* Python version: 3.8.5

### Description

Creating a numpy masked array with the mask attribute set to something other than the default value of False raises AttributeError

### What I Did

```python
>>> import numpy as np
>>> from unyt import s
>>> np.ma.MaskedArray(range(2)*s, mask=[True, True])
...
/.../unyt/unyt/array.py in __getitem__(self, item)
   1625     def __getitem__(self, item):
   1626         ret = super(unyt_array, self).__getitem__(item)
-> 1627         if ret.shape == ():
   1628             return unyt_quantity(
   1629                 ret, self.units, bypass_validation=True, name=self.name

AttributeError: '_MaskedPrintOption' object has no attribute 'shape'
```

l-johnston:
Closes #169

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1624,10 +1624,12 @@ class unyt_array(np.ndarray):
 
     def __getitem__(self, item):
         ret = super(unyt_array, self).__getitem__(item)
-        if hasattr(ret, "shape") and ret.shape == ():
-            ret = unyt_quantity(ret, self.units, bypass_validation=True, name=self.name)
-        elif hasattr(ret, "units"):
-            ret.units = self.units
+        if getattr(ret, "shape", None) == ():
+            ret = unyt_quantity(ret, bypass_validation=True, name=self.name)
+        try:
+            setattr(ret, "units", self.units)
+        except AttributeError:
+            pass
         return ret
 
     #
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1624,14 +1624,11 @@ class unyt_array(np.ndarray):
 
     def __getitem__(self, item):
         ret = super(unyt_array, self).__getitem__(item)
-        if ret.shape == ():
-            return unyt_quantity(
-                ret, self.units, bypass_validation=True, name=self.name
-            )
-        else:
-            if hasattr(self, "units"):
-                ret.units = self.units
-            return ret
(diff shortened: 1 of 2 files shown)


l-johnston:
@ngoldbaum I ran the `pyperf` [benchmark](https://gist.github.com/l-johnston/75293a7b072d37c6629c58de8a36fce5) and get the following results:

- v2.7.2: 2.83 µs (SD=0.11 µs)
- proposed: 2.97 µs (SD=0.16 µs)

l-johnston:
@ngoldbaum Is there anything else to do?

ngoldbaum:
I haven’t had time to carefully look at this yet, please have some patience.

neutrinoceros (approved):
I have played with this a little and this fits very well into my (admittedly limited) usage of masked arrays. I don't know how to seriously evaluate the performance cost of this patch, but I'll add that all my attempts to refactor the logic failed, so I can get behind what's been done.
@jzuhone might still want to have a look.

jzuhone:
It would be very nice to figure out the performance impact first.

l-johnston:
@jzuhone I have measured performance previously:

> @ngoldbaum I ran the `pyperf` [benchmark](https://gist.github.com/l-johnston/75293a7b072d37c6629c58de8a36fce5) and get the following results:
> 
> * v1.0.1: 3.01 µs (SD=0.11 µs) (benchmark version for paper)
> * v2.7.2: 2.83 µs (SD=0.11 µs)
> * proposed: 2.97 µs (SD=0.16 µs)

is this what you are looking for?

neutrinoceros:
@l-johnston thanks. I'm now running a modified version of you benchmark with a much heavier array

neutrinoceros:
So I ran your benchmark with a 3GiB array, each run took 10min on my machine and this is what I'm getting out of it:


current master: 
Mean +- std dev: 1.26 us +- 0.09 us

this patch:
Mean +- std dev: 1.26 us +- 0.17 us

so the standard deviation is definitely affected, but the raw performance not so much, as far as I can tell.

PR #172, opened 2020-09-09
Fix unyt_array.to() to support complex-valued arrays

### LINKED ISSUE #171: unyt_array.in_units() casts complex numbers to real losing imaginary part
* unyt version: 2.7.2
* Python version: 3.8.5

### Description
Attempted to convert unit of a complex valued array.

### What I Did

```
>>> from unyt import Ω, F, Hz
>>> freq = [1j, 1j*10]*Hz
>>> arr = 1/(Ω*F*freq)
>>> arr.to("dimensionless")
.../numpy/core/_asarray.py:83:ComplexWarning: Casting complex values to real discards the imaginary part
return array(a, dtype, copy=False, order=order)
unyt_array([0., 0.], dtype=float128, units='dimensionless')
```

The problem is in array.py, `unyt_array.in_units` line 846:
```
new_dtype = np.dtype("f" + str(dsize))
```

As a workaround, I can use `unyt_array.in_base()` and then check for `dimensionless`.

l-johnston:
Closes #171


ngoldbaum:
It looks like the matplotlib code changed but none of the tests changed to reflect that, can you add tests if that’s appropriate? I’d also be more comfortable merging this if the tests did more to make sure that the library works correctly with conplex dtypes (e.g. testing more than just `to`).

l-johnston:
The latest version of black (20.8b1) caused formatting changes in the docstring. I did manual testing of the other conversion functions and they work fine, but I will add them to the unit test.

unyt/array.py
```diff
@@ -843,21 +843,14 @@ def in_units(self, units, equivalence=None, **kwargs):
                         RuntimeWarning,
                         stacklevel=2,
                     )
-            new_dtype = np.dtype("f" + str(dsize))
-            conversion_factor = new_dtype.type(conversion_factor)
-            ret = np.asarray(self.ndview * conversion_factor, dtype=new_dtype)
```
jzuhone, line 848: Is there a reason this got switched from `ndview` to `v`? The latter involves making a copy, which we want to avoid if we can.

unyt/array.py
```diff
@@ -982,8 +975,6 @@ def in_base(self, unit_system=None):
         else:
             to_units = self.units.get_base_equivalent(unit_system)
             conv, offset = self.units.get_conversion_factor(to_units, self.dtype)
-        new_dtype = np.dtype("f" + str(self.dtype.itemsize))
```
jzuhone, line 985: Why doesn't this require the same logic as above, instead of just getting rid of these two lines?

unyt/array.py
```diff
@@ -843,21 +843,14 @@ def in_units(self, units, equivalence=None, **kwargs):
...
                         stacklevel=2,
                     )
-            new_dtype = np.dtype("f" + str(dsize))
-            conversion_factor = new_dtype.type(conversion_factor)
-            ret = np.asarray(self.ndview * conversion_factor, dtype=new_dtype)
+            new_dtypekind = "c" if self.dtype.kind == "c" else "f"
+            new_dtype = np.dtype(new_dtypekind + str(dsize))
+            ret = np.asarray(self.v * conversion_factor, dtype=new_dtype)
             if offset:
-                np.subtract(ret, offset, ret)
```
jzuhone, line 850: It's not clear to me why these lines needed to be changed here.

jzuhone (changes requested):

unyt/array.py
l-johnston: Inadvertent change while prototyping the fix. I find that `ndview` works fine and will switch back to it.

unyt/array.py
l-johnston: I seem to remember having trouble with the `bypass_validation=True` during the prototyping process. Reverting these lines appears to be acceptable.

unyt/array.py
l-johnston: Because `in_base` doesn't explicitly create a new array like `in_units` does. The conversion factor `conv` has the correct dtype and numpy casting rules produce the desired result.

jzuhone (approved):
I'm satisfied with this now. @ngoldbaum?

PR #176, opened 2020-10-09
Fix SI prefix scaling of degree Celsius

### LINKED ISSUE #175: Temperature unit degree Celcius with SI prefixes has wrong value.
* unyt version: v2.7.2
* Python version: 3.8.6

### Description

NIST SP811 permits the use of SI prefixes with degree Celcius, but `unyt` gives the wrong value.

### What I Did

```
>>> from unyt.testing import allclose_units
>>> from unyt import mdegC, degC, K
>>> (1*mdegC).in_base()
unyt_quantity(0.27415, 'K')
>>> (0.001*degC).in_base()
unyt_quantity(273.151, 'K')
>>>allclose_units(1*mdegC, 0.001*degC)
False
```

l-johnston:
Closes #175

### DIFF
--- a/unyt/tests/test_units.py
+++ b/unyt/tests/test_units.py
@@ -858,6 +858,16 @@ def test_degC():
     assert str(a) == "1 °C"
 
 
+def test_degC_with_SIprefixes():
+    assert_allclose_units(1 * Unit("mdegC"), 0.001 * Unit("degC"))
+    assert_allclose_units(1 * Unit("degC"), 1000 * Unit("mdegC"))
+    assert_allclose_units(73 * Unit("degF"), 22777.779 * Unit("mdegC"))
+    assert_allclose_units(22777.779 * Unit("mdegC"), 73 * Unit("degF"))
+    assert_allclose_units(22777.779 * Unit("mdegC"), 532.67 * Unit("R"))
+    assert_allclose_units(1 * Unit("mK"), -273149.0 * Unit("mdegC"))
+    assert_allclose_units(1 * Unit("mdegC"), 273151.0 * Unit("mK"))
+
+
 def test_delta_degC():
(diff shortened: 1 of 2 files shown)


jzuhone:
Can you check performance for this change? I want to make sure there's not a significant regression.

ngoldbaum:
I’m pretty sure it won’t matter much for `unyt_array` since the unit operations are behind a cache.

ngoldbaum:
Doing this with string interpolations feels super janky, I’m pretty sure there’s another way to do this that doesn’t require getting the conversion factors from the repr.

l-johnston:
@ngoldbaum I agree the solution is less than ideal. I have spent quite a few hours studying the issue and yes it is possible to obtain the conversion factors at a higher level, but the trouble is distinguishing `degC`, and variants, from `degF` to then know how to manipulate the conversion factors. For example,

``` 
>>> mdegC.get_conversion_factor(K)
(0.001, -0.27315)
>>> degF.get_conversion_factor(K)
(0.5555555555555556, -255.37222222222223)
```

Visually, it is obvious what to do, but in code there's no attribute I am aware of that distinguishes `mdegC` from `degF`. They're both in the same "mks" registry and have the same base_equivalent of kelvin. There's no attribute indicating a prefixed unit. The registry has a list of prefixable units, but these are strings.

I chose this approach as being the least disruptive to the design.

ngoldbaum:
Why not add the necessary metadata to the unit object somehow?

I’m happy to look at this at some point if you can’t come up with a better solution.

l-johnston:
@ngoldbaum I switched to using `_split_prefix` to determine when to modify the conversion factors. Is this approach acceptable?

jzuhone:
@l-johnston this approach looks good to me.

jzuhone (approved):

PR #180, opened 2020-11-15
Fix #179 by handling matplotlib case for limits

### LINKED ISSUE #179: Matplotlib's unyt support breaks when plotting zero-length arrays
* unyt version: unyt==2.7.1
* Python version: numpy==1.19.0 python==3.8.0, matplotlib==3.2.2
* Operating System: MacOS and Linux

### Description

When using `matplotlib_support`, if we plot an empty unyt array, matplotlib breaks when changing the units. See the below case.

We have people interested in helping fix this but don't know if it's an unyt problem or a matplotlib problem.

### What I Did

The breaking case:
```python
import unyt
import matplotlib.pyplot as plt
import numpy as np

x = unyt.unyt_array([1, 2, 3], units="Mpc")
y = unyt.unyt_array([1, 2, 3], units="Solar_Mass")

mask = np.array([False, False, False])

x_lims = unyt.unyt_array([0, 1], units="Mpc")

 with unyt.matplotlib_support:
     plt.plot(x[mask], y[mask])
     plt.xlim(*x_lims)
```
Full traceback:
```
AttributeError                            Traceback (most recent call last)
<ipython-input-10-5f09c5659620> in <module>
      1 with unyt.matplotlib_support:
      2     plt.plot(x[mask], y[mask])
----> 3     plt.xlim(*x_lims)
      4 

~/.pyenv/versions/3.8.0/lib/python3.8/site-packages/matplotlib/pyplot.py in xlim(*args, **kwargs)
   1394     if not args and not kwargs:
   1395         return ax.get_xlim()
-> 1396     ret = ax.set_xlim(*args, **kwargs)
   1397     return ret
   1398 

~/.pyenv/versions/3.8.0/lib/python3.8/site-packages/matplotlib/axes/_base.py in set_xlim(self, left, right, emit, auto, xmin, xmax)
   3181             right = xmax
   3182 
-> 3183         self._process_unit_info(xdata=(left, right))
   3184         left = self._validate_converted_limits(left, self.convert_xunits)
   3185         right = self._validate_converted_limits(right, self.convert_xunits)

~/.pyenv/versions/3.8.0/lib/python3.8/site-packages/matplotlib/axes/_base.py in _process_unit_info(self, xdata, ydata, kwargs)
   2071             return kwargs
   2072 
-> 2073         kwargs = _process_single_axis(xdata, self.xaxis, 'xunits', kwargs)
   2074         kwargs = _process_single_axis(ydata, self.yaxis, 'yunits', kwargs)
   2075         return kwargs

~/.pyenv/versions/3.8.0/lib/python3.8/site-packages/matplotlib/axes/_base.py in _process_single_axis(data, axis, unit_name, kwargs)
   2054                 # We only need to update if there is nothing set yet.
   2055                 if not axis.have_units():
-> 2056                     axis.update_units(data)
   2057 
   2058             # Check for units in the kwargs, and if present update axis

~/.pyenv/versions/3.8.0/lib/python3.8/site-packages/matplotlib/axis.py in update_units(self, data)
   1514         neednew = self.converter != converter
   1515         self.converter = converter
-> 1516         default = self.converter.default_units(data, self)
   1517         if default is not None and self.units is None:
   1518             self.set_units(default)

~/.pyenv/versions/3.8.0/lib/python3.8/site-packages/unyt/mpl_interface.py in default_units(x, axis)
    108             # label in the subsequent call to axisinfo
    109             unyt_arrayConverter._axisnames[axis] = name if name is not None else ""
--> 110             return x.units
    111 
    112         @staticmethod

AttributeError: 'tuple' object has no attribute 'units'
```

JBorrow:
Fixes #179 by handling the case where a tuple is passed to `default_units` (which occurs when the plots is empty and a limit call is made). I've also added a test.

Perhaps @l-johnston would like to take a look at this.

### DIFF
--- a/unyt/mpl_interface.py
+++ b/unyt/mpl_interface.py
@@ -102,12 +102,20 @@ else:
 
             Unit object
             """
-            name = getattr(x, "name", "")
+            # In the case where the first matplotlib command is setting limits,
+            # x may be a tuple of length two (with the same units).
+            if isinstance(x, tuple):
+                name = getattr(x[0], "name", "")
+                units = x[0].units
+            else:
+                name = getattr(x, "name", "")
+                units = x.units
+
             # maintain a mapping between Axis and name since Axis does not point to
             # its underlying data and we want to propagate the name to the axis
             # label in the subsequent call to axisinfo
             unyt_arrayConverter._axisnames[axis] = name if name is not None else ""
-            return x.units
+            return units
 
         @staticmethod
         def convert(value, unit, axis):
--- a/unyt/tests/test_mpl_interface.py
+++ b/unyt/tests/test_mpl_interface.py
@@ -234,3 +234,9 @@ def test_multiple_subplots():
     assert generated_labels == expected_labels
     _matplotlib.pyplot.close()
     matplotlib_support.disable()
+
+
+@check_matplotlib
+def test_empty_plot(ax):
+    ax.set_xlim(*unyt_array([0, 1], "s"))
+    _matplotlib.pyplot.close()


ngoldbaum:
LGTM. 

The test failures aren’t your fault, I think, something to do with how we’re using hdf5 in the doc examples, something probably changed upstream.

I need to move the tests to github actions because of travis’ new build time policies and will fix that then. I will try to fix the tests soon so you aren’t blocked here. It’s probably ok to merge this with the failing tests

JBorrow:
Thanks, I think you're right about the test failures - this would only affect the MPL support tests which all pass (including the new one, of course).

neutrinoceros:
Hey @JBorrow, CI was migrated to GitHub Actions. Can you rebase this please ?

neutrinoceros (approved):
Ok I was able to check that the added test is red on master and green here, and I confirm that the patch fixes the issue addressed. I think we can simply disregard the two "❌" from old CI system since I could obtain validation by closing and reopening.
@jzuhone, can we merge this ? :)

PR #184, opened 2021-03-08
MNT: Cfradial units addition for PR https://github.com/yt-project/yt/pull/1990

zssherman:
Hello I recently was pinged for working on getting the [PR](https://github.com/yt-project/yt/pull/1990) cfradial support into yt. Currently working on fixing the PR now. I saw the units are now imported and added it here. I'm not honestly sure if anything else is needed to be added as compared to how the original units were coded back in 2018, any help would be appreciated and I can commit and fix if need be. Thanks for your time!

### DIFF
--- a/unyt/_unit_lookup_table.py
+++ b/unyt/_unit_lookup_table.py
@@ -261,7 +261,6 @@ default_unit_symbol_lut = OrderedDict(
         ("counts", (1.0, dimensions.dimensionless, 0.0, r"\rm{counts}", False)),
         ("photons", (1.0, dimensions.dimensionless, 0.0, r"\rm{photons}", False)),
         ("dBZ", (1.0, dimensions.dimensionless, 0.0, r"\rm{dBZ}", False)),
-        ("dB", (1.0, dimensions.dimensionless, 0.0, r"\rm{dB}", False)),
         ("me", (mass_electron_kg, dimensions.mass, 0.0, r"m_e", False)),
         ("mp", (mass_hydrogen_kg, dimensions.mass, 0.0, r"m_p", False)),
         ("Sv", (1.0, dimensions.specific_energy, 0.0, r"\rm{Sv}", True)),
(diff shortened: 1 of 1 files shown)


neutrinoceros:
This looks reasonable to me. @zssherman would you be able to rebase this so it can be properly validated by CI ?

zssherman:
Sounds good @neutrinoceros, did the rebase and repushed.

zssherman:
Slightly confused on the duplicate db error.

neutrinoceros:
Got it. That's because we already define `"B"` (bel) here https://github.com/yt-project/unyt/blob/a65c04cc2d198bae90c3aa77acfaa9dd7f12b2ff/unyt/_unit_lookup_table.py#L477
so `"dB"` is automatically registered too.
I think you can simply remove it.

zssherman:
@neutrinoceros Ah gotcha, sounds good, will remove db definition.

neutrinoceros:
Ok so this looks fine to me but I must say I am unsure what's the policy for adding domain specific units.
If the new Yt frontend requires this change it'd be a good idea to include it in next release.
Pinging @jzuhone for a more authoritative opinion.

jzuhone:
So in general if a unit is in fairly common use, we'd just add it here in unyt. I'm not necessarily opposed to doing that here. 

However, it may make more sense to add it in the cfradial frontend. When you `load` a dataset into yt you can add units to the unit registry for that particular frontend at that time. The FITS frontend adds some of its own units here, for example:

https://github.com/yt-project/yt/blob/e76b39b8f8ed24c972e8acc1b250a5f5bbbc89a4/yt/frontends/fits/data_structures.py#L685-L699

Now of course this is a slightly different case because we don't know what `beam` or `pixel` will be until the datasets are loaded in the FITS case, whereas here `dBZ` is a well-defined thing independent of a dataset. But unless I'm mistaken it's only really used in this context and so it may make sense to define it from within the frontend in a similar way.

jzuhone:
For more context, you could do it in the cfradial frontend by overriding `_set_code_unit_attributes` (if you haven't already), call the version from the `super`, and then do:

```python
self.unit_registry.add("dBZ", 1.0, dimensions=dimensions.dimensionless)
```

zssherman:
@jzuhone Gotcha, makes sense. Ill add both units the the unit registry in the frontend data structures and push the commit to the YT PR. Thanks for the input!

neutrinoceros:
Alright, sounds to me that we can close this one then.

PR #185, opened 2021-04-14
daskified unyt arrays

chrishavlin:
This PR introduces the `unyt_dask_array` class, which implements a subclass of standard dask arrays with units attached. Still a work in progress, but it is generally useable now!

Basic usage (also shown here in a [notebook](https://nbviewer.jupyter.org/github/chrishavlin/yt_scratch/blob/master/notebooks/dask_unyt_array_updates.ipynb)) begins by using the `unyt_from_dask` function to create a new `unyt_dask_array` instance from a dask array:

```python
from unyt.dask_array import unyt_from_dask
from dask import array as dask_array
x = unyt_from_dask(dask_array.random.random((10000,10000), chunks=(1000,1000)), 'm')
x
Out[2]:  unyt_dask_array<random_sample, shape=(10000, 10000), dtype=float64, chunksize=(1000, 1000), chunktype=numpy.ndarray, units=m>
```

The array can be manipulated as any other dask array:

```python
result = (x * 2).mean()
result
Out[3]: unyt_dask_array<mean_agg-aggregate, shape=(), dtype=float64, chunksize=(), chunktype=numpy.ndarray, units=m>
result.compute()
Out[4]:  unyt_quantity(1.00009275, 'm')
```

If the return is an array, we get a `unyt_array` instead:

```python
(x * 2 + x.to('cm')).mean(1).compute()
Out[8]: unyt_array([1.50646938, 1.48487083, 1.49774744, ..., 1.49939197,
            1.49462512, 1.48263323], 'm')
```

Unit conversions:

```python
x = unyt_from_dask(dask_array.random.random((10000,10000), chunks=(1000,1000)), 'lb')
x.mean().compute()
Out[9]:
    unyt_quantity(0.50002619, 'lb')
x.in_mks().mean().compute()
Out[10]: unyt_quantity(0.22680806, 'kg')
x.to('mg').mean().compute()
Out[11]: unyt_quantity(226808.06379903, 'mg')
from unyt import g
x.to(g).mean().compute()
Out[12]: unyt_quantity(226.8080638, 'g')
```

The implementation relies primarily on decorators and a hidden `unyt_dask_array._unyt_array` to track unit conversions and has very minimal modifications to the existing unyt codebase.  If a user is running a dask client, then all the above calculations will be executed by that client (see [notebook](https://nbviewer.jupyter.org/github/chrishavlin/yt_scratch/blob/master/notebooks/dask_unyt_array_updates.ipynb)), but the implementation here only needs the dask array subset (i.e., `pip install dask[array]`). 

### Some remaining known issues:
- [x] reductions return standard dask arrays when using external functions (see note below)
- [x] dask is added to `_on_demand_imports` but haven't added it to the testing environment yet, so new tests will fail
- [x] haven't yet done flake8/etc checks
- [x] no new docs yet (update: added to the usage page)
- [x] new tests could use a bit more tweaking
- [x] squash commits? I have a lot... but would be easy to squash. let me know. (update: chose not to squash)

Note on the issue with dask reductions: 

If you do:

```python
from unyt.dask_array import unyt_from_dask
from dask import array as dask_array

x = unyt_from_dask(dask_array.random.random((10000,10000), chunks=(1000,1000)), 'm')
x.min().compute()
```
You get a `unyt_quantity` as expected: `unyt_quantity(0.50002407, 'm')`

But if you use the daskified equivalent of `np.min(ndarray)`:

```python
dask_array.min(x).compute()
```
You get a plain float: `0.50002407`. This isn't much of an issue for simple functions like `min`, but many more complex functions are not implemented as attributes. Not yet sure what the best approach is here...

Update (8/24) to the dask reductions: I've played around with many approaches focused around manually wrapping all of the dask reductions but have decided that the added complexity is not worth it. Instead, I added a user-facing function, `unyt.dask_array.reduce_with_units` that accepts a dask function handle, the unyt array and any `args` and `kwargs`  for the dask function that internally wraps the dask function handle to track units. 

### standalone package?

One final note: while I've been developing this to be incorporated into unyt, the fact that there are very minimal changes to the rest of the codebase means that this **could** be a standalone package. Happy to go that route if it seems more appropriate!


ngoldbaum:
Just a heads up that the tests are currently broken and need to be migrated to github actions. So far I haven’t summoned the will to figure out github actions. Any help on that front from anyone reading this is much appreciated.

chrishavlin:
latest commits include a rebase, some formatting fixes, improvements (slicing now returns unyt-dask arrays rather than plain dask arrays).

neutrinoceros:
Hey @chrishavlin, CI migration is now done, maybe you'd like to rebase this branch to validate it ?

chrishavlin:
Some performance testing results!

The following 2 plots are the median and minimum computation time as a function of array size for standard numpy arrays, standard unyt arrays, standard dask arrays and the new dask-unyt arrays for an element-wise and reduction operation: `(x ** 2).mean()`. 20 iterations were run for all test cases at each array size. In the case of the dask and dask-unyt arrays, the chunksize is fixed at 1e7 for all array sizes. Additionally, the dask and dask-unyt cases were run with two different local cluster configurations of 4 and 6 workers (one thread per worker in both cases). 

fixed_chunk_median
fixed_chunk_min

Results are conceptually similar to those I showed at scipy ([link](https://chrishavlin.github.io/scipy2021/01_unytdask.html)) -- the unyt-dask array sees a slight decrease in performance compared to a standard dask array (presumably because of all the decorating) but it's a small fraction compared to the performance gain when comparing the unyt-dask array to standard unyt arrays at larger array sizes. Note that I ran these on my desktop rather than laptop (as for scipy) so the cross-over point where the dask and unyt-dask arrays exhibit shorter computation times is at a smaller array size and computation scales much better at larger array sizes (i.e., the slope of the dask and unyt-dask curves is smaller here than what I showed at scipy). 

Additionally, I did some tests for just the dask and unyt-dask arrays varying chunksize at a fixed array size of 1e9. Same test computation as before, using 4 workers and 1 thread per worker. The following image shows the median computation time over 20 iterations at each chunk size: 

chunk_dependence_median

Similar conclusion as the fixed chunksize results: there is a minimal amount of overhead for the unyt-dask array compared to a standard dask array. 

Happy to run some more comparisons if anyone has ideas for specific test cases, just let me know!

jzuhone (approved):

chrishavlin:
thanks for reviewing, @jzuhone ! I know I still need to write some documentation for this -- I'm thinking it'll probably be more straightforward as its own PR. If that sounds good I'll open an issue to remind myself to write it.

matthewturk:
@chrishavlin Do you want some help with this, or do you think the documentation is doable?

chrishavlin:
Re: documentation, I can certainly write up some docs. Should I just add a dask section to the `docs/usage.rst` page?

jzuhone:
I'm ok with merging this now or including docs with it, whichever you prefer, @chrishavlin

neutrinoceros:
If I understand correctly, this PR is required for https://github.com/yt-project/yt/pull/3511. Doesn't it mean we'll need a new unyt release too ?

chrishavlin:
@neutrinoceros at some point ya, but I can also split https://github.com/yt-project/yt/pull/3511 . The dask-unyt array functionality there is only for returning the dask-unyt arrays to the user, the actual read functionality does not require dask-unyt arrays.  

@jzuhone given that I forgot about writing up the docs here, I'm going to just add to this PR so that I make myself do it in a timely manner. I added it to my to-do list for this week! So updates soon!

chrishavlin:
Just added some docs -- will look into the test failures, don't think they're related to my changes, might be a tox issue. but more importantly I also want to check if #200  could actually simplify alot of what I've done here... so don't merge quite yet.

chrishavlin:
Currently failing test is not related to this PR: https://github.com/yt-project/unyt/pull/219#issuecomment-1048663750

neutrinoceros:
@chrishavlin CI is stable again, could you update this to fix the couple existing conflicts ?

chrishavlin:
great! will do!

chrishavlin:
Looks like a few more minor fixes needed after that merge... should be straightforward.

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+    return False
+
+
+def _is_iterable(obj):
+    return type(obj) is tuple or type(obj) is list
+
+
+def _sanitize_unit_args(*input):
+    # returns sanitized inputs and unyt_inputs for calling the ufunc
+    # what if i is a list or tuple????? np.concatenate([x,x2]) is one arg...
```
chrishavlin, line 156: reading back through my changes... will look into this comment I left myself...

neutrinoceros: ok then don't hesitate to ping me when it's ready. I can review this later this week, I think

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+        return _post_ufunc(super()._elemwise, unyt_result)(ufunc, *args, **kwargs)
+
+    def __array_ufunc__(self, numpy_ufunc, method, *inputs, **kwargs):
+        inputs, unyt_result = _prep_ufunc(
+            numpy_ufunc, *inputs, extract_dask=True, **kwargs
+        )
+        wrapped_func = _post_ufunc(super().__array_ufunc__, unyt_result)
+        return wrapped_func(numpy_ufunc, method, *inputs, **kwargs)
+
+    def __array_function__(self, func, types, args, kwargs):
```
neutrinoceros, line 305: We should really make sure there are no flagrant incompatibilities with #200. 
I think #200 is more fundamental so I would argue it should go in first if possible, but I know it's a big one too. At a minimum, we should try to run all tests in a temporary merger branch.

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+
+def unyt_from_dask(
+    dask_array,
+    units=None,
+    registry=None,
+    bypass_validation=False,
+    unyt_name=None,
+):
+    """
+    creates a unyt_dask_array from a standard dask array.
```
neutrinoceros, line 502: The docstring should clarify whether a copy of the raw data is made or not.

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+# x_da = unyt_from_dask(dask.array.ones((10, 10), chunks=(2, 2)), unyt.m)
+# dask.array.min(x_da).compute()  #  returns a plain float
+# x_da.min().compute()  #  returns a unyt quantity
+#
+# but when the functions do not exist as attributes, like dask.array.nanmin(),
+# it is difficult to handle without manually wrapping all of those reductions
+# functions and exposing them here. The following function, reduce_with_units,
+# is a compromise: it is a simple helper function for calling
+# a general dask.array.reductions method on a unyt_dask array to correctly
+# handle units.
```
neutrinoceros, line 566: very useful comment, I would suggest to convert it to an issue and link it here, so there's a better chance of finding it back in the future (IMO)

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+        # e.g., min, max, nanstd. functions that return the same units can
+        # be called directly as the dask function will treat unyt_dask_in as
+        # a standard dask array and then we copy over the initial units.
+        return _simple_dask_decorator(dask_func, unyt_dask_in)(
+            unyt_dask_in, *args, **kwargs
+        )
+    else:
+        # the operation may change the units
+        npfunc = getattr(np, dask_func.__name__, None)
+        if npfunc:
```
neutrinoceros, line 614: just a personal style preference. I may be blinded by the fact that we recently gained access to the walrus operator in unyt by dropping support for Python 3.7, so please ignore this suggestion if you think it hurts readability
```suggestion
        if npfunc := getattr(np, dask_func.__name__, None)
```

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+    out._set_unit_state(units, new_unyt_array, unyt_name)
+    return out
+
+
+def unyt_from_dask(
+    dask_array,
+    units=None,
+    registry=None,
+    bypass_validation=False,
+    unyt_name=None,
```
neutrinoceros, line 499:
```suggestion
    *,
    registry=None,
    bypass_validation=False,
    unyt_name=None,
```
see https://github.com/yt-project/yt/issues/3992 if you need an explanation/rationale

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+        name,
+        chunks,
+        dtype=None,
+        meta=None,
+        shape=None,
+        units=None,
+        registry=None,
+        bypass_validation=False,
+        input_units=None,
+        unyt_name=None,
```
neutrinoceros, line 257: my interpretation of the docs is that this class isn't intended as publicly exposed, but it feels safer to treat it as if it was.
```suggestion
        *,
        dtype=None,
        meta=None,
        shape=None,
        units=None,
        registry=None,
        bypass_validation=False,
        input_units=None,
        unyt_name=None,
```
Also, it seems that a lot of these arguments are not re-exposed via `unyt_from_dask`, which I understand is the actual public interface. Should they be removed or exposed ?

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+    def __getitem__(self, index):
+        return _simple_dask_decorator(super().__getitem__, self)(index)
+
+    def __setitem__(self, key, value):
+        # requires dask >= 2021.4.1
+        super().__setitem__(key, value)
+
+    def __repr__(self):
+        disp_str = super().__repr__().replace("dask.array", "unyt_dask_array")
+        units_str = f", units={self.units.__str__()}>"
```
neutrinoceros, line 320: I don't think this makes a difference ?
```suggestion
        units_str = f", units={self.units}>"
```

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+
+    def _repr_html_(self):
+        # controls jupyter notebook display of an array. called from _repr_html_
+        base_table = super()._repr_html_()
+        table = base_table.split("\n")
+        new_table = []
+        for row in table:
+            if "</tbody>" in row:
+                u = self.units.__str__()
+                newrow = f"    <tr><th> Units </th><td> {u} </td> <td> {u} </td></tr>"
```
neutrinoceros, line 331: same as above
```suggestion
                newrow = f"    <tr><th> Units </th><td> {u} </td> <td> {self.units} </td></tr>"
```

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+    return wrapper
+
+
+# the following list are the unyt_quantity/array attributes that will also
+# be attributes of the unyt_dask_array. These methods are wrapped with a unit
+# conversion decorator.
+_unyt_funcs_to_track = ["to", "in_units", "in_cgs", "in_base", "in_mks"]
+
+
+def _track_conversion(unyt_func_name, current_unyt_dask):
```
neutrinoceros, line 77: the way this function is used suggests it should be a method

neutrinoceros (commented):
Overall great work ! here are a bunch of comments, suggestions and questions.

neutrinoceros:
@chrishavlin I checked that merging this with #200 didn't cause any of their respective tests to fail so we can push forward with this PR now.
I'm ready to approve and merge this if can resolve my threads with a quick dialogue.

docs/usage.rst
```diff
@@ -1377,3 +1377,52 @@ There are three ways to use the context manager:
...
+
+Methods that hang off of a :mod:`unyt_dask_array` object and operations on
+:mod:`unyt_dask_array` objects will generally preserve units:
+
+    >>> x_da.sum().compute()
+    unyt_quantity(49995000, 'm')
+    >>> (x_da[:5000] * x_da[5000:]).compute()[:5]
+    unyt_array([    0,  5001, 10004, 15009, 20016], 'm**2')
+
+One important caveat is that using Dask array functions may strip units:
```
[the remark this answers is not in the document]
chrishavlin: Yes and no... I have to double check, but I'm pretty sure the issue was that the dask array functions end up pulling out arrays for each chunk (so a unyt array ends up using just the `.value` and units get lost). But there may be a way to deal with that using the `__array_function__` protocol and dask chunk types... but I was never able to find a solution that wouldn't also require re-writing or wrapping even more of the dask array functionality. There very is likely a way to do it though. Maybe coming back to this now with fresh eyes will help. I'll take another look at it after I get through the remaining comments!

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+__doctest_requires__ = {
+    ("dask_array.unyt_from_dask"): ["dask"],
+}
+
+
+_dask_Array = dask.array.core.Array
+_dask_finalize = dask.array.core.finalize
+
+# the following attributes hang off of dask.array.core.Array and do not modify units
+_use_simple_decorator = [
```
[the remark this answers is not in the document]
chrishavlin: cool, can do that easily enough.... will come back to this.

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+        units=None,
+        registry=None,
+        bypass_validation=False,
+        input_units=None,
+        unyt_name=None,
+    ):
+
+        # get the base dask array
+        obj = super(unyt_dask_array, clss).__new__(
+            clss,
```
chrishavlin, line 262:
```suggestion
            cls,
```

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+        >>> from unyt import dask_array
+        >>> import dask.array as da
+        >>> x = da.random.random((10000, 10000), chunks=(1000, 1000))
+        >>> x_da = dask_array.unyt_from_dask(x, 'm')
+        >>> x_da.to_dask()
+        dask.array<random_sample, shape=(10000, 10000), dtype=float64,
+             chunksize=(1000, 1000), chunktype=numpy.ndarray>
+        """
+        (cls, args) = self.__reduce__()
+        return super().__new__(_dask_Array, *args)
```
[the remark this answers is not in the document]
chrishavlin: will look into this, but I'm pretty sure there was a reason for calling `__new__` directly... will figure that out!

unyt/dask_array.py
chrishavlin: will have to double check if there a reason not too, but looking now I do think all these kwargs should be in `unyt_from_dask` as well.

unyt/tests/test_dask_arrays.py
```diff
@@ -0,0 +1,411 @@
...
+    _use_simple_decorator,
+    reduce_with_units,
+    unyt_dask_array,
+    unyt_from_dask,
+)
+from unyt.exceptions import UnitOperationError
+from unyt.unit_symbols import cm, g, m
+
+
+def dask_is_too_old(dask_version_str):
```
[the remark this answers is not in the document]
chrishavlin: release notes are here https://docs.dask.org/en/stable/changelog.html#v2021-04-1 but not sure it's actually available anymore anywhere. in any case, I think just requiring a newer version is reasonable.

unyt/tests/test_dask_arrays.py
```diff
@@ -0,0 +1,411 @@
...
+
+def test_unyt_dask_slice():
+
+    # tests __getitem__
+    x = dask.array.ones((10, 10), chunks=(2, 2))
+    x_da = unyt_from_dask(x, m)
+    slc = x_da[:, 0]
+    assert slc.units == m
+    assert slc.compute().units == m
+    assert type(slc.compute()) == unyt_array
```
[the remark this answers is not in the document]
chrishavlin: will go through and update all the others too!

unyt/tests/test_dask_arrays.py
```diff
@@ -0,0 +1,411 @@
...
+    unary_result_test(x_da.min(), x_unyt.min())
+    unary_result_test(x_da.max(), x_unyt.max())
+    unary_result_test(x_da.mean(), x_unyt.mean())
+    unary_result_test(x_da.std(), x_unyt.std())
+    unary_result_test(x_da.cumsum(0), x_unyt.cumsum(0))
+    unary_result_test(abs(x_da), abs(x_unyt))  # __abs__
+
+
+def test_logical():
+    def logical_operators(x_da, other):
```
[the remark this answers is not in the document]
chrishavlin: yes... yes it could... pretty sure I wrote this before I much of anything about pytest parameterizing... would be nice to use it tho! will do so.

docs/usage.rst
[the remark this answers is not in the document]
neutrinoceros: FTR I mostly meant to raise concerns with how this PR interplays with #200, but in the mean time I've also checked that all of their tests combined still pass if we merge them together.

unyt/dask_array.py
neutrinoceros: Update: I checked and found no real merge conflict and 0 test failures

unyt/dask_array.py
chrishavlin: now a method! I think it was only a function because of how development evolved... but in its current form it does make more sense as a class method. which is now is.

unyt/dask_array.py
chrishavlin: updated (it makes a copy of the dask task graph, not raw data)

unyt/dask_array.py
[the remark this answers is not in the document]
chrishavlin: changed to a set (along with a few other lists that were used similarly)

unyt/dask_array.py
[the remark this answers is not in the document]
chrishavlin: well, couldn't figure out why I had been calling new directly, works fine with your suggestion.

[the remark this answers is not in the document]
chrishavlin: (the code is now as you suggested)

unyt/dask_array.py
```diff
@@ -0,0 +1,620 @@
...
+]
+
+
+def _simple_dask_decorator(dask_func, current_unyt_dask):
+    """
+    a decorator for the simpler dask functions that can just copy over the current
+    unit info after applying the dask function. this includes functions that return
+    single values (e.g., min()) or functions that change the array in ways that do
+    not affect units like reshaping or rounding.
+
```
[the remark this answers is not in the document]
chrishavlin: yup, I agree. renamed.

unyt/tests/test_dask_arrays.py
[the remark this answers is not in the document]
chrishavlin: bumped the minimum version for the tests

unyt/tests/test_dask_arrays.py
```diff
@@ -0,0 +1,411 @@
...
+        assert_array_equal(result, actual)
+
+
+def test_bad_dask_array_reductions():
+    x_da = unyt_from_dask(dask.array.ones((10, 10), chunks=(2, 2)), m)
+
+    def empty_func():
+        pass
+
+    with pytest.raises(ValueError):
```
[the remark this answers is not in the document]
chrishavlin: done

unyt/tests/test_dask_arrays.py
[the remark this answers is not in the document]
chrishavlin: now parametrized!

chrishavlin:
oh, oops. looks like the minimum dask version for testing caused some problems with 3.8. Let me relax that a little...

chrishavlin:
also, I realized that pickling won't actually work quite right... I spent some time trying thinking that it'd be trivial, but it's not. At this point I'd rather leave it to a followup (will open an issue in a moment)

chrishavlin:
Ok, @neutrinoceros I think the current version addressed all your comments. And I've opened issues to track the long comment/note on dask reductions as you suggested as well as an issue to track the pickle issue. 

IMO the pickle issue isn't worth holding up a merge, but if anyone disagrees I'll of course defer. In any case, I'm planning to take a closer look in the next few days -- I'm hoping that it is actually an easy fix and I just wasn't seeing it in my initial attempt today...

chrishavlin:
Oh wait -- I think the one unresolved thing here is whether #250 should go in before this? I'm fine if that's the case.

neutrinoceros (approved):
> IMO the pickle issue isn't worth holding up a merge, but if anyone disagrees I'll of course defer. In any case, I'm planning to take a closer look in the next few days

I agree; we can disclose the issue in the next release notes if necessary.


> I think the one unresolved thing here is whether https://github.com/yt-project/unyt/pull/250 should go in before this?

Not necessarily, I can include the change in my PR too if this one goes in first.

I would merge immediately but I'm going to hold off for the day in case there's anything left to discuss that I missed, or in case you want to make another attempt to fix the pickling issue.
If nothing comes up in the next 24h I'll just merge :)

chrishavlin:
figured out the pickling! turns out I just needed to read the pickle docs again rather than try to do it from memory... will push up by the end of today.

PR #188, opened 2021-05-25
fix a bug where units would be printed twice in formatted strings with a unyt_array embedded

neutrinoceros:
This bug was discovered in a discussion with @chummels
```python
>>> import unyt as u
>>> d = u.unyt_array([1, 2, 3], "Msun")
>>> print("%s"%d) # ok
[1 2 3] Msun
>>> print("{}".format(d)) # broken
[1 2 3] Msun Msun
>>> print(f"{d}") # broken too
[1 2 3] Msun Msun
```

Weirdly, `numpy.ndarray.__format__` apparently already includes the units in a `str.format` call. I hope this patch fixes the issue in a Numpy/Python version-agnostic way.
Note that I used a f-string in the test, though I'm not sure that the minimal tested Python version is >=3.6.
The bug was however also reproducible by calling `str.format` directly so this part of the test is not essential.


unyt/array.py
```diff
@@ -589,8 +589,7 @@ def __str__(self):
         return str(self.view(np.ndarray)) + " " + str(self.units)
 
     def __format__(self, format_spec):
-        ret = super(unyt_array, self).__format__(format_spec)
-        return ret + " {}".format(self.units)
+        return "{} {}".format(self.v.__format__(format_spec), self.units)
```
jzuhone, line 592: Use `self.d` here because `self.v` makes a copy.

jzuhone (changes requested):

unyt/array.py
neutrinoceros: good call. Fixed !

jzuhone:
it appears the test failures are unrelated

neutrinoceros:
@jzuhone anything I can do to help fixing CI, even if the cause is external to this PR ?

jzuhone:
@neutrinoceros I thought that was being handled here: https://github.com/yt-project/unyt/pull/190

neutrinoceros:
@jzuhone this is now rebased and ready for review :)

jzuhone (approved):

jzuhone:
@matthewturk should have a quick look here

matthewturk (approved):
LGTM

PR #190, opened 2021-05-27
ci: Migrate CI to GitHub Actions workflows

matthewfeickert:
* Closes PR #187
* Requires PR #189

This PR migrates CI from Travis CI and Appveyor to use GitHub Actions workflows. The GHA CI will run across Ubuntu, MacOS, and Windows environments across CPython runtimes spanning 3.6 to 3.9. To reduce the number of runs (especially on slower runners like MacOS) the test matrix only runs on MacOS and Windows for the edge CPython versions: Python 3.6 and Python 3.9. The CI runs on a variety of event triggers:
- Pushes to the `master` branch (PR merges trigger "push" events)
- Pushes to pull requests
- As a nightly CRON job (useful for being alerted to dependencies breaking APIs)
- On demand manual triggers

Travis CI and Appveyor are dropped in this PR and coverage reporting is switched over to use the Codecov GHA (this will require some follow up from the maintainers as you'll want to get an optional `CODECOV_TOKEN` to _greatly_ speed up reporting).

PR #194, opened 2021-08-09
MAINT: Drop Python 3.5 support

matthewfeickert:
* Should follow PR #190

This PR follows up on PR #174 and drops Python 3.5 support and additionally updates the docs and PyPI metadata to reflect this. As part of that docs updates it also adds Python 3.9 as a supported version.


versioneer.py
```diff
@@ -10,7 +10,7 @@
 * https://github.com/warner/python-versioneer
 * Brian Warner
 * License: Public Domain
-* Compatible With: python2.6, 2.7, 3.2, 3.3, 3.4, 3.5, 3.6, and pypy
+* Compatible With: python3.6, 3.7, 3.8, 3.9, and pypy
```
neutrinoceros, line 13: while we're at it
```suggestion
* Compatible With: python3.6, 3.7, 3.8, 3.9, and PyPy
```

neutrinoceros (approved):
LGTM

neutrinoceros (changes requested):
sorry about the churn.
Also, I think you need to rebase this to remove the appveyor conf file

versioneer.py
```diff
@@ -10,7 +10,7 @@
 * https://github.com/warner/python-versioneer
 * Brian Warner
 * License: Public Domain
-* Compatible With: python2.6, 2.7, 3.2, 3.3, 3.4, 3.5, 3.6, and pypy
+* Compatible With: python3.6, 3.7, 3.8, 3.9, and PyPy
```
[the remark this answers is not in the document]
matthewfeickert: I've undone this in the interactive rebase that I'm about to push.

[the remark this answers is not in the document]
neutrinoceros: excellent, thank you

neutrinoceros (approved):

setup.py
```diff
@@ -48,7 +49,7 @@
     packages=find_packages(include=["unyt"]),
     test_suite="tests",
     tests_require=test_requirements,
-    python_requires=">=3.5",
+    python_requires=">=3.6",
```
neutrinoceros, line 52: Cargo culting a bit but I've been seeing a lot of packages requiring 3.6.1, instead of 3.6.0
I've been less and less motivated to figure out why though, since Python 3.6 is soon to be EOL

neutrinoceros (approved):

setup.py
matthewfeickert: Easy enough to do and I don't think anyone is going to complain on that. :)

jzuhone (approved):

jzuhone:
@ngoldbaum I don't know if you want to weigh in on this one or not

neutrinoceros:
We probably want to merge this before we can add explicit support for Python 3.10, which we should be able to do in a few weeks.
@ngoldbaum I believe you already signed off to drop Python 3.5 (though I can't remember where), are you okay with this PR ?

PR #195, opened 2021-08-10
MNT: upgrade versioneer install to 0.20

neutrinoceros:
Upgrade performed with
```shell
python -m pip install --upgrade versioneer
versioneer install
```
Rationale: the newest version of versioneer has explicit support for Python 3.6 to 3.9


neutrinoceros:
This PR is already "rotted" in a sense, now we probably should have explicit support for Python 3.10 as well, but it's not even clear that the version currently installed is really broken, so I'll just close this.

PR #196, opened 2021-08-11
FEAT: add unyt.testing_assert_equal_units

neutrinoceros:
In https://github.com/yt-project/yt/pull/3463, it became apparent that in some cases `numpy.testing.assert_equal` wasn't enough to validate that a unyt_array instance wasn't changed (it doesn't validate that units stayed constant)
This patch implements `unyt.testing.assert_equal_units` in line with `unyt.testing.assert_allclose_units`.


neutrinoceros:
making this a draft for now. I'd like to carefully inspect that the docstrings (essentially copy-pasted) still make sense, but I have to run.

neutrinoceros:
While working on this I found an annoying bug that I think comes from numpy where error messages from array comparison functions (like `numpy.testing.assert_array_equal`) omit the unit when representing the compared array, leaving a confusing error message in a failing test, e.g.,
```
E       AssertionError:
E       Not equal to tolerance rtol=1e-07, atol=0
E
E       Mismatched elements: 1 / 3 (33.3%)
E       Max absolute difference: 1.
E       Max relative difference: 0.25
E        x: unyt_array([1, 2, 3])
E        y: unyt_array([1.e-05, 2.e-05, 4.e-05])
```
I'll remove the draft status when I'm able to confirm the bug isn't in unyt or when I have a fix for it.

neutrinoceros:
fell down a rabbit hole. I'm now pretty convinced that the interface to fix the issue I mentioned above is implemented as part of [NEP 18](https://numpy.org/neps/nep-0018-array-function-protocol.html).

neutrinoceros:
confirmed that the missing part can be implemented following NEP 18. This is done in #200

Issue #139, opened 2020-02-10
__array_function__ support


ngoldbaum:
I spent a little bit of time this weekend working on adding `__array_function__` to `unyt_array`. The work is on the `array-function` [branch on my fork](https://github.com/ngoldbaum/unyt/tree/array-function).

Unfortunately I think it might take quite a bit of effort to get a working implementation. This is because right now if a type defines `__array_function__` then you signal to numpy that your implementation will work for all numpy functions that support `__array_function__` (afaics pretty much the whole public API, although I can't find a public listing of all functions I'd need to wrap so I need to double-check that by inspecting the numpy source code). If you implement `__array_function__` but do not have wrappers for a function you can either return `NotImplemented` for that function, in which case users will see a `TypeError` when they try to use that function, or you can attempt to coerce the input data to `ndarray`. The latter is really what I'd like, since that would preserve backward compatibility, however it's very difficult because the APIs of all of the functions that support `__array_function__` are so heterogenous, so there isn't a single easy heuristic one can apply to coerce arguments to `ndarray`.

@l-johnston if you'd like to take a crack at this starting from my branch please feel free. I'll comment here if I go back to working on this and systematically wrapping the whole numpy API.

If we do add `__array_function__` support I think the resulting product would have to be `unyt 3.0`, it's too much of a change in API for downstream users for it not to be. The fact that we're changing the types of what comes out of operations in downstream code is also very concerning, I'm not sure what breakage would happen. We probably need to test downstream projects that depend on `unyt` to make sure that adding `__array_function__` support doesn't cause a lot of breakage.

shoyer:
Falling back to coercing to `ndarray` is _possible_ though not necessarily advisable. Dask does something like this in its `__array_function__` method. The main thing you need is some sort of "nested coercion" function that handles nested containers of your array type.

In practice, I think something like the following could work pretty well:
```python
def nested_coerce(args):
    if isinstance(args, dict):
        return {k: nested_coerce(v) for k, v in args.items()}
    elif isinstance(args, tuple):
        return tuple(nested_coerce(a) for a in args)
    elif isinstance(args, list):
        return [nested_coerce(a) for a in args]
    elif isinstance(args, MyArray):
        return np.asarray(args)
    else:
        return args
```
You can all this like `args, kwargs = nested_coerce((args, kwargs))` inside your `__array_function__` method. It won't handle completely arbitrary containers, but generally those aren't valid for use with NumPy, anyways.

[...]

Issue #130, opened 2020-01-26
np.histogram does not support unyt_array


l-johnston:
* unyt version: 2.5.0
* Python version: 3.6.9
* Operating System: Linux

### Description

Computing a histogram where both the input data and range parameters are both of unyt_array type crashes. See [numpy issue 15437](https://github.com/numpy/numpy/issues/15437) for discussion there.

### What I Did

```
>>> import numpy as np
>>> from unyt import unyt_array
>>> x = unyt_array(np.random.normal(size=1000), "m")
>>> np.histogram(x, bins=10, range=(x.min(), x.max()))
```

This issue causes matplotlib's histogram to fail - even without matplotlib_support enabled.

```
>>> import matplotlib.pyplot as plt
>>> from unyt import m
>>> import numpy as np
>>> x = np.random.normal(size=1000)*m
>>> plt.hist(x)
```

ngoldbaum:
For what it’s worth I’m planning on taking a first pass on implementing `__array_function__` for `unyt_array`, that said it’s probably not correct for numpy to error like that for operands that do not implement `__array_function__`.

Issue #69, opened 2019-03-08
Using np.linalg.inv with unyt_array has no effect on units


sharkweek:
* unyt version: 1.0.7
* Python version: 3.6
* Operating System: Windows 10

### Description

Using the `numpy.linalg.inv()` to calculate the inverse of a matrix, doesn't affect units.

### What I Did
Here is what happens when you take the inverse of an `unyt_array`:
```
>>> x = unyt.unyt_array([[1. , 0.5, 0.5],
                         [0.5, 1. , 0.5],
                         [0.5, 0.5, 1. ]],
                        'psi')
>>> np.linalg.inv(x)
unyt_array([[ 1.5, -0.5, -0.5],
            [-0.5,  1.5, -0.5],
            [-0.5, -0.5,  1.5]], 'psi')
```
`numpy` correctly carries out the inverting operation, but the units remain the same where they should now be the reciprocal of the original units (`'inch**2/lbf'` in this case). As of now, the only work around I can think of is to create a new `unyt_array` with the inverted value of the original one and the reciprocal of the original's units.

ngoldbaum:
So I doubt there's anything we can do about making `np.lingalg.inv` respect units correctly without support in NumPy for `__array_function__`, which is being added in a future numpy release. I don't see any mention of `lingalg.inv` in [NEP-18](https://www.numpy.org/neps/nep-0018-array-function-protocol.html) but maybe @shoyer knows more?

We could also define a wrapper around `lingalg.inv` that correctly handles units and add the wrapper to `unyt`.

shoyer:
Every function that accepts arrays or array-like objects in NumPy's public APIs will support NEP 18. That includes `np.linalg.inv`.

sharkweek:
That's good news. I suppose we just have to wait until the numpy release that implements `__array_function__`.

Issue #50, opened 2018-08-15
Inconsistent behaviour between np.arange and np.linspace


dangom:
* unyt version: 1.0.6
* Python version: 3.6.6
* Operating System: MacOS 10.13.6

### Description
When using `np.arange` on quantities, I'd expect the result to be a unyt_array, just like when using `np.linspace`. The result, however, is a standard numpy array.

### What I Did

```
In [310]: type(np.linspace(5*s,6*s,100))
Out[310]: unyt.array.unyt_array
In [311]: type(np.arange(5*s,6*s,100*u.ms))
Out[311]: numpy.ndarray
```

ngoldbaum:
I doubt there’s anything we can do about this without defining a wrapper around arange that behaves as you’re expecting. It may be the case that in the future we could override this with `__array_function__` which is being actively discussed on NumPy’s github.

Juanlu001:
Reference: https://github.com/numpy/numpy/issues/12028 and https://www.numpy.org/neps/nep-0018-array-function-protocol.html

shoyer:
`np.linspace` is documented as accepting scalars fro `start`/`stop` rather than arrays/array-like. So following the rule that `__array_function__` is only called on array-like arguments, it currently (on master) it is not slated for `__array_function__` support. But perhaps this is worth revisiting?

ngoldbaum:
Yet another case where arrays-with-units are special: we want scalars with units to work seamlessly too :)

saulshanabrook:
I am not sure if this is the right place to bring this up, but it would great to be able to define custom dispatching for function like `np.arange` or `np.empty`. 

I guess one option would be for to NumPy to dispatch on scalars as well as arrays. Is this incompatible with the NEP? Would anything break with this change?

ngoldbaum:
I'm pretty sure that's a question for the numpy issue tracker or mailing list, not for the unyt bug tracker :)

saulshanabrook:
@ngoldbaum Sorry! I was clicking around on issues and didn't realize I had switched repos :)

PR #200, opened 2021-08-19
ENH: Provisional support for NEP 18 (__array_function__ protocol)

neutrinoceros:
My initial motivation here was to add some unit representation to error messages when comparing two unyt_array instance via functions from `numpy.testing`, like so:

```python
import numpy as np
import unyt as un

a = [1, 2, 3] * un.cm
b = [1, 2, 3] * un.km
np.testing.assert_array_equal(a, b)
```
which yields, on master:

```
...
AssertionError:
Arrays are not equal

Mismatched elements: 3 / 3 (100%)
Max absolute difference: 299997.
Max relative difference: 0.99999
 x: unyt_array([1, 2, 3])
 y: unyt_array([1, 2, 3])
```
and on this branch
<details>
<summary>previous version</summary>

```
...
AssertionError:
Arrays are not equal

Mismatched elements: 3 / 3 (100%)
Max absolute difference: 299997. cm
Max relative difference: 0.99999 dimensionless
 x: unyt_array([1, 2, 3] cm)
 y: unyt_array([1, 2, 3] km)
```

</details>

edit:
```
AssertionError:
Arrays are not equal

Mismatched elements: 3 / 3 (100%)
Max absolute difference: 299997., units='cm'
Max relative difference: 0.99999, units='dimensionless'
 x: unyt_array([1, 2, 3], units='cm')
 y: unyt_array([1, 2, 3], units='km')
```

Incidentally, it turns out that fixing this necessitated a kick-off implementation of NEP 18, so this work laid the foundation to solve:
- [x] #69
- [x] #130
- [ ] #50 (most likely out of scope)

More broadly, implementing NEP 18 for any was the topic of #139.
Granted I need to take more time to check that I'm not going against the original intentions here. My current approach is that, since covering the whole numpy public API in one go seems like an gigantic task, I'm implementing `unyt_array.__array_function__` with a fallthrough condition: if a special case isn't implemented yet, just fallback to the raw numpy implem (which is currently the behaviour for _all_ functions subject to NEP 18). This way we can add support for more and more functions in a progressive way.
I'm going to set the bar low(ish) for now, and try and fix the already reported cases, as reported above, as a first step.

An important question to address is: what should be done in the general case where we don't have a custom implementation for an array function ?

1) transparently default to the raw numpy implementation without a warning (this is de facto what is done as of unyt 2.8.0, and will remain the case until NEP 18 isn't at least partially implemented).
2) same as 1. _but_ emit a warning (possibly, we could have a whitelist of functions that are known to be perfectly fine without a specialised implementation, for which no warning would be emitted) along the lines of 
```
UserWarning: calling `numpy.FUNC` on a unyt_array. Results may hold incorrect units. A future version of unyt will remove this warning, and possibly change the behaviour of this function to be dimensionally correct.
```
3) Error out

Option 1 this is the current implementation of this PR because I think it is the less disruptive or noisy one. My personal opinion is that it's probably okay to have incomplete support for NEP 18 for a couple release, as long as it is clearly stated in the release notes.

### DIFF
--- a/unyt/_deprecation.py
+++ b/unyt/_deprecation.py
@@ -1,0 +1,22 @@
+import warnings
+from typing import Optional
+
+
+def warn_deprecated(
+    name,
+    /,
+    *,
+    stacklevel: int = 3,
+    replacement: Optional[str] = None,
+    since_version: str,
+) -> None:
+    msg = (
+        f"{name} is deprecated and will be removed in a future version\n"
+        f"Instead, {replacement}\n"
+        f"(deprecated since unyt v{since_version})"
+    )
+    warnings.warn(
+        msg,
+        category=DeprecationWarning,
+        stacklevel=stacklevel,
+    )
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -2064,7 +2064,8 @@ def test_numpy_wrappers():
     assert_array_equal(unorm(a3, axis=1), unyt_array(arr_norm_answer, "cm"))
     assert_array_equal(np.linalg.norm(a3, axis=1), arr_norm_answer)
 
-    assert_array_equal(udot(a1, a1), unyt_quantity(dot_answer, "cm**2"))
+    with pytest.warns(DeprecationWarning):
+        assert_array_equal(udot(a1, a1), unyt_quantity(dot_answer, "cm**2"))
 
     assert_array_equal(np.array(catenate_answer), uconcatenate((a1.v, a2.v)))
     with pytest.raises(RuntimeError):
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -1,4 +1,5 @@
 import numpy as np
+
 from unyt.exceptions import UnitConversionError
 
 _HANDLED_FUNCTIONS = {}
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -1,0 +1,18 @@
+import numpy as np
+
+_HANDLED_FUNCTIONS = {}
+
+
+def implements(numpy_function):
+    """Register an __array_function__ implementation for unyt_array objects."""
+    # See NEP 18 https://numpy.org/neps/nep-0018-array-function-protocol.html
+    def decorator(func):
+        _HANDLED_FUNCTIONS[numpy_function] = func
+        return func
+
+    return decorator
+
+
+@implements(np.array2string)
+def array2string(a, *args, **kwargs):
+    return np.array2string._implementation(a, *args, **kwargs) + f" {a.units}"
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -16,7 +16,10 @@ def implements(numpy_function):
 
 @implements(np.array2string)
 def array2string(a, *args, **kwargs):
-    return np.array2string._implementation(a, *args, **kwargs) + f" {a.units}"
+    return (
+        np.array2string._implementation(a, *args, **kwargs)
+        + f", units={str(a.units)!r}"
+    )
 
 
 def _get_conversion_factor(out, units) -> float:
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -20,7 +20,18 @@ def array2string(a, *args, **kwargs):
 
 @implements(np.linalg.inv)
 def linalg_inv(a, *args, **kwargs):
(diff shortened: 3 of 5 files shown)


neutrinoceros:
I've gone the long way and started implemented a bunch of functions with obvious requirements.
TODO:

## inverse-like functions
- [x] `np.linalg.inv`
- [x] `np.linalg.tensorinv`
- [ ] `np.linalg.pinv` (done-ish, more testing needed for the rcond parameter)

## product-like functions
- [x] `np.dot`
- [ ] `np.inner`
- [ ] `np.outer`
- [ ] `np.matmul`

## histogram functions (as requested for #130)
- [x] `np.histogram`
- [x] `np.histogram2d`
- [x] `np.histogramdd`

neutrinoceros:
pre-commit.ci autofix

neutrinoceros:
@jzuhone do you think you'd be able to review this ? Is there someone else I could ask ?

jzuhone:
@neutrinoceros I can try to look at it next week

neutrinoceros:
@jzuhone I would like to propose we bundle this PR in the next release alongside the dask layer. IMO, each of these would justify a release on their own, so it's up to you :)

jzuhone:
@chrishavlin or @matthewturk could you look at this one?

ngoldbaum:
I'm a little concerned that the coverage of numpy functions isn't as extensive as users might like and that implementing support in a piecemeal fashion might lead to more breakage than users would prefer, since fundamentally we're changing the type of the outputs of numpy functions. If we're going to break that, it would be best to break them all at once instead of staggering that over many releases as support gets added in a piecemeal fashion.

unyt ships with a number of wrappers for numpy functions (`uconcatenate`, `ucross`, `udot`, etc). We should make sure we have replacements for all of those (since there's clearly already a need for those) and deprecate the wrappers. I know at least yt uses those wrappers.

If you want a hand with this I'm happy to help out.

neutrinoceros:
I agree that it'd be best to do it all in one clean go, and yes I would love a hand here. How should we proceed ? If you want to keep it all in a single PR feel free to push to my branch.

ngoldbaum:
Sure I’ll push to this branch

neutrinoceros:
@ngoldbaum do you know of a reliable way to generate the complete list of functions we should be covering ?

ngoldbaum:
Yes, see https://github.com/numpy/numpy/issues/15544#issuecomment-584453408 as well as other comments in that issue

ngoldbaum:
Can you make sure that the existing wrappers in unyt that this PR obsoletes (e.g. `udot`) are deprecated? You could probably rewrite them to be thin wrappers around the regular numpy functions now.

neutrinoceros:
good call, I completely missed that. I think it may be preferable _not_ to rewrite them at all in case there's _any_ unanticipated difference so users have a chance to report and still get something that works for them while we fix the replacement, but I'll definitely add those deprecation warnings.

neutrinoceros:
I rebased the branch but didn't write the history. I added two commit, the first of which corresponds to #288, and the second one adds deprecation warnings as requested.

neutrinoceros:
pre-commit.ci autofix

neutrinoceros:
(rebased again now that #288 went in)

unyt/tests/test_unyt_array.py
```diff
@@ -39,6 +31,15 @@
...
 
 from unyt import K, R, Unit, degC, degF, delta_degC, delta_degF, dimensions
+
+# ----------------------------------------------------------------------------
+# Copyright (c) 2013, yt Development Team.
+#
+# Distributed under the terms of the Modified BSD License.
+#
+# The full license is in the file COPYING.txt, distributed with this software.
+# ----------------------------------------------------------------------------
```
ngoldbaum, line 41: Can you move this header back to the top of the file?

unyt/_deprecation.py
```diff
@@ -0,0 +1,26 @@
+import warnings
+from typing import Optional
+
+
+class UnytDeprecationWarning(UserWarning):
+    pass
```
ngoldbaum, line 6: Can we use the built-in `DeprecationWarning` instead? We already use that elsewhere.

ngoldbaum (commented):
Just a couple minor nits from the changes you made today. Will merge once they're resolved.

unyt/tests/test_unyt_array.py
neutrinoceros: oh, good catch. I think this was done by `isort` or something, but I should have seen it

PR #202, opened 2021-09-06
BUG: expose a bug with typing for unyt_quantity

### LINKED ISSUE #201: BUG: unyt_quantity are `Sized` objects without a length ?
* unyt version: 2.8.0
* Python version: 3.9.6
* Operating System: OS/X

### Description

`unyt_quantity` objects are recognised as instances of `collections.abc.Sized`, which indicates that they have a `__len__` protocol. However, calling `len(quantity)` fails with an error message contradicting this.

### What I Did

```python
from collections.abc import Sized
import unyt as un

quan = un.unyt_quantity(1, "cm")
assert isinstance(quan, Sized)
print(len(quan))
```
output
```
TypeError: len() of unsized object
```

neutrinoceros:
endgoal: fix #201

Right now I'm just adding a failing test.

### DIFF
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -2563,3 +2563,11 @@ def test_invalid_unit_quantity_from_string(s):
         match="Could not find unit symbol '{}' in the provided symbols.".format(un_str),
     ):
         unyt_quantity.from_string(s)
+
+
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Closing since the problem is actually on numpy's side.

PR #207, opened 2021-10-20
MNT: add explicit support for Python 3.10

neutrinoceros:
follow up to #194
This will likely fail CI at first, it may be a little early to expect support is already provided by unyt's dependencies, so I'll open as a draft for now, and see what happens.


.github/workflows/ci.yml
```diff
@@ -18,17 +18,27 @@ jobs:
...
+          - os: macos-latest
+            python-version: '3.9'
+          - os: windows-latest
+            python-version: '3.7'
           - os: windows-latest
-            python-version: 3.7
+            python-version: '3.8'
           - os: windows-latest
-            python-version: 3.8
+            python-version: '3.9'
```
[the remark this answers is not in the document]
neutrinoceros: if I use this, wouldn't I need to keep a growing list of includes for ubuntu ?
Also, I'm not sure the list is excepted to grow much more actually. We'll probably want to drop support for 3.6 once it either reaches EOL or when it starts making maintenance harder after EOL.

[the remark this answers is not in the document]
matthewfeickert: > if I use this, wouldn't I need to keep a growing list of includes for ubuntu?

No, you still just use the `python-version` list c.f. https://github.com/jiffyclub/snakeviz/blob/33b1452cb91958f802cb8f8efc2d336cb23fadd9/.github/workflows/ci.yml#L16-L27 which is

```yaml
jobs:
  test:
    runs-on: ${{ matrix.os }}
    strategy:
      matrix:
        os: [ubuntu-latest]
        python-version: ["2.7", "3.5", "3.6", "3.7", "3.8", "3.9", "3.10"]
        include:
          - os: macos-latest
            python-version: "3.7"
          - os: macos-latest
            python-version: "3.10"
```

Also, I'm not sure the list is excepted to grow much more actually. We'll probably want to drop support for 3.6 once it either reaches EOL or when it starts making maintenance harder after EOL.

> I'm not sure the list is excepted to grow much more actually.

Sure, I expect that too. To my mind the use of `include` makes more sense, which is why I've adopted it everywhere else.

> We'll probably want to drop support for 3.6 once it either reaches EOL

I think 3.6 is basically there already. https://www.python.org/dev/peps/pep-0494/ Python 3.6.15 [notes that](https://www.python.org/downloads/release/python-3615/)

>  Python 3.6 is now in the security fix phase of its life cycle. Only security-related issues are accepted and addressed during this phase. We plan to provide security fixes for Python 3.6 as needed through 2021, five years following its initial release.

So there's less than 2 months until EOL.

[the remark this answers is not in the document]
neutrinoceros: If you're fine with it I'm on board !

neutrinoceros:
h5py 3.6 was released a couple days ago with wheels for Python 3.10 and was the main offender. Let's try to rerun CI to see if we can get away with building astropy on the fly

neutrinoceros:
this is now failing because of a regression in Matplotlib, reported as #208

neutrinoceros:
Everything is ready upstream, and CI is green, so I'm opening this for review. @ngoldbaum, you have my blessing if you want/need to update anything here to make it compatible with #214

jzuhone:
@neutrinoceros is this still ready? can you rebase on top of the latest master?

neutrinoceros:
@jzuhone done !

jzuhone:
@neutrinoceros what's the relationship between this and PR #214? ping @ngoldbaum also

neutrinoceros:
@jzuhone rebased !

CONTRIBUTING.rst
```diff
@@ -95,48 +95,48 @@ managing your python evironment using your operating system's package manager or
     $ export PYENV_ROOT="$HOME/.pyenv"
     $ export PATH="$HOME/.pyenv/bin:$PATH
     $ eval "$(pyenv init -)"
-    $ pyenv install -s 3.6.14
-    $ pyenv install -s 3.7.11
-    $ pyenv install -s 3.8.11
-    $ pyenv install -s 3.9.6
+    $ pyenv install -s 3.7.12
```
jzuhone, line 98: We don't need this line anymore, correct?

CONTRIBUTING.rst
```diff
@@ -95,48 +95,48 @@ managing your python evironment using your operating system's package manager or
...
-5. When you're done making changes, check that your changes pass ``flake8``,
+7. When you're done making changes, check that your changes pass ``flake8``,
    format the code with ``black``, and run the tests, including testing several
    Python versions with ``tox``::
 
     $ flake8 unyt
     $ black ./
     $ pytest --doctest-modules --doctest-rst --doctest-plus
-    $ pyenv local 3.6.14 3.7.11 3.8.11 3.9.6
+    $ pyenv local 3.7.13 3.8.13 3.9.12 3.10.4
```
jzuhone, line 125: Or this one

CONTRIBUTING.rst
```diff
@@ -164,7 +164,7 @@ execute ``tox`` in the root of the ``unyt`` repository. For example, using the
 ``pyenv`` environment we set up above::
 
    $ cd unyt
-   $ pyenv local 3.6.14 3.7.11 3.8.11 3.9.6
+   $ pyenv local 3.7.13 3.8.13 3.9.12 3.10.4
```
jzuhone, line 167: Here too

CONTRIBUTING.rst
neutrinoceros: good call. Cleaned it as well as other remaining occurrences

neutrinoceros:
Need to bump minimal supported h5py to 3.0.0 to get windows wheels too

jzuhone (approved):

Issue #208, opened 2021-11-18
Regression in matplotlib interface with matplotlib 3.5


neutrinoceros:
### Description

Matplotlib 3.5 (just released) apparently breaks unit's interface.

### What I Did

```
$ pytest unyt/tests/test_mpl_interface.py::test_errorbar
...
unyt.exceptions.UnitOperationError: The <ufunc 'add'> operator for unyt_arrays with units "kg" (dimensions "(mass)") and "dimensionless" (dimensions "1") is not well defined.
```
This test pass with matplotlib 3.4.3 but fails with matplotlib 3.5
It's likely a regression on the matplotlib side but I wanted to report here to keep track of the problem.

neutrinoceros:
reported upstream as https://github.com/matplotlib/matplotlib/issues/21669

neutrinoceros:
So right now I'm not sure where the problem should be addressed, but what seems clear is that we could have caught this 6 months earlier, had we tested against matplotlib's dev branch. I'm not blaming anyone for this, but we can do better in the future, especially now that we're using GitHub Actions, so I'll open a PR addressing this aspect.

neutrinoceros:
This was resolved upstream with https://github.com/matplotlib/matplotlib/pull/21884

PR #209, opened 2021-11-18
TST: run daily CI job against dev branches of matplotlib and numpy

neutrinoceros:
Motivated by #208

### DIFF
--- a/.github/workflows/bleeding-edge.yaml
+++ b/.github/workflows/bleeding-edge.yaml
@@ -1,0 +1,48 @@
+name: CI (bleeding edge)
+# this workflow is heavily inspired from pandas, see
+# https://github.com/pandas-dev/pandas/blob/master/.github/workflows/python-dev.yml
+
+# gcheck stability against dev version of Python, numpy, and matplotlib
+
+on:
+  push:
+    branches:
+      - master
+  # Run daily at 0:01 UTC
+  schedule:
+    - cron:  '1 0 * * *'
+  workflow_dispatch:
+
+jobs:
+  build:
+    runs-on: ubuntu-latest
+    name: Python3.10-dev
+    timeout-minutes: 60
+
+    concurrency:
+      group: ${{ github.ref }}-dev
+      cancel-in-progress: true
+
+    steps:
+    - uses: actions/checkout@v2
+      with:
+        fetch-depth: 0
+
+    - name: Set up Python Dev Version
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.10-dev'
+
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade setuptools wheel
+        python3 -m pip install git+https://github.com/numpy/numpy.git
+        python3 -m pip install git+https://github.com/matplotlib/matplotlib.git
+        python3 -m pip install pytest
+
+    - name: Build unyt
+      run: python3 -m pip install --no-build-isolation .
+
+    - name: Run Tests
+      run: pytest -vvv unyt/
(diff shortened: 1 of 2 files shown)


neutrinoceros:
running the first iteration of this new workflow here: https://github.com/neutrinoceros/unyt/actions/runs/1477213382
So currently it is expected to _fail_, but only because of #208. If it fails for any other reason, I'll iterate.

neutrinoceros:
Got it https://github.com/neutrinoceros/unyt/runs/4255286000?check_suite_focus=true
ping @ngoldbaum for review :)

ngoldbaum:
Sorry I can’t really help with github actions stuff.

neutrinoceros:
This one is based on a battle tested workflow I setup on yt and that we've used for months.
@matthewturk could maybe give his benediction?

matthewturk (commented):
Looks OK to me, but I'm not really in any position to provide a "benediction" -- @jzuhone @Xarthisius ?

ngoldbaum:
I rebased this on master to get CI to pass. I'll go ahead and merge it now, please send followups if you notice issues with the actual cron jobs. I'm not sure who gets notifications if it fails.

neutrinoceros:
Sure thing. From past events with yt I expect to get notified myself but even if not, I'llbe watching over this for the next couple weeks, or as long as necessary. Thanks for merging!

.github/workflows/bleeding-edge.yaml
```diff
@@ -0,0 +1,48 @@
...
+    - name: Set up Python Dev Version
+      uses: actions/setup-python@v2
+      with:
+        python-version: '3.10-dev'
+
+    - name: Install dependencies
+      run: |
+        python3 -m pip install --upgrade pip
+        python3 -m pip install --upgrade setuptools wheel
+        python3 -m pip install git+https://github.com/numpy/numpy.git
```
matthewfeickert, line 40: Hi @neutrinoceros @ngoldbaum. This is a belated comment but while doing some looking at the CI for PR #214 I noticed this was created and thought I'd note that we were told not do this (building NumPy or SciPy from source) in https://github.com/scikit-hep/pyhf/pull/1533 by Ralf and asked to use the nightly wheels instead. c.f https://anaconda.org/scipy-wheels-nightly/repo

.github/workflows/ci.yml
```diff
@@ -5,9 +5,6 @@ on:
     branches:
     - master
   pull_request:
-  # Run daily at 0:01 UTC
-  schedule:
-  - cron:  '1 0 * * *'
```
matthewfeickert, line 10: why was this removed? You want to know if your current versions break against released dependencies even if `--pre` builds are breaking/passing.

neutrinoceros: I'm essentially assuming that if we get to catch upstream breakage ahead of time, we should never see anything break with releases, but that may be too optimistic of an assumption. I wouldn't oppose restoring this.

.github/workflows/bleeding-edge.yaml
neutrinoceros: I don't understand why we'd want to take one step further away from the source since it's trivial to build numpy ourselves. The time it takes (about 20min) seems irrelevant given this job isn't run on PRs. What am I missing ?

ngoldbaum: Because users of numpy don’t normally build it from source so it’s a more realistic test of the distribution.

.github/workflows/ci.yml
ngoldbaum: I disagree with this logic, I’m gonna add it back.

PR #211, opened 2021-11-18
support coercing iterables with nonuniform dimensionally equivalent units to a single unyt_array

### LINKED ISSUE #210: Cannot coerce N-tuple of `unyt_array` instances with different units into a single N-dimensional `unyt_array` instance.
* unyt version: 2.8.0
* Python version: 3.7.3

### What I Did

```
In [1]: import unyt as u

In [2]: d = ([1, 2]*u.g, [1, 2]*u.kg)

In [3]: u.unyt_array(d)
IterableUnitCoercionError                 Traceback (most recent call last)
<ipython-input-3-8c7b43202134> in <module>
----> 1 u.unyt_array(d)

~/Documents/unyt/unyt/array.py in unyt_array.__new__(cls, input_array, units, registry, dtype, bypass_validation, input_units, name)
    551 elif _iterable(input_array) and input_array:
    552     if isinstance(input_array[0], unyt_array):
--> 553         return _coerce_iterable_units(input_array, registry)
    555 # Input array is an already formed ndarray instance
    556 # We first cast to be our class type
    558 obj = np.asarray(input_array, dtype=dtype).view(cls)

~/Documents/unyt/unyt/array.py in _coerce_iterable_units(input_object, registry)
    246 ff = getattr(input_object[0], "units", NULL_UNIT)
    247 if any([ff != getattr(_, "units", NULL_UNIT) for _ in input_object]):
--> 248     raise IterableUnitCoercionError(input_object)
    249 # This will create a copy of the data in the iterable.
    250 return unyt_array(np.array(input_object), ff, registry=registry)

<class 'str'>: (<class 'TypeError'>, TypeError('not all arguments converted during string formatting'))
```

The error is also confusing because the error we're supposed to be raising is itself raising a `TypeError` when we try to instantiate it (d'oh!).

Astropy handles this case just fine:

```
In [1]: import astropy.units as u

In [2]: d = ([1, 2]*u.g, [1, 2]*u.kg)

In [3]: u.Quantity(d)
Out[3]:
<Quantity [[1.e+00, 2.e+00],
           [1.e+03, 2.e+03]] g>
```

ngoldbaum:
Fixes #210.

See the test I modified for a usage example.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -257,7 +257,13 @@ def _coerce_iterable_units(input_object, registry=None):
         if any([isinstance(o, unyt_array) for o in input_object]):
             ff = getattr(input_object[0], "units", NULL_UNIT)
             if any([ff != getattr(_, "units", NULL_UNIT) for _ in input_object]):
-                raise IterableUnitCoercionError(input_object)
+                ret = []
+                for datum in input_object:
+                    try:
+                        ret.append(datum.in_units(ff.units))
+                    except UnitConversionError:
+                        raise IterableUnitCoercionError(str(input_object))
+                return unyt_array(np.array(ret), ff, registry=registry)
             # This will create a copy of the data in the iterable.
             return unyt_array(np.array(input_object), ff, registry=registry)
     return np.asarray(input_object)
--- a/unyt/exceptions.py
+++ b/unyt/exceptions.py
@@ -161,15 +161,15 @@ class IterableUnitCoercionError(Exception):
     Example
     -------
 
-    >>> from unyt import km, cm, unyt_array
-    >>> data = [2*cm, 3*km]
+    >>> from unyt import g, cm, unyt_array
+    >>> data = [2*cm, 3*g]
     >>> unyt_array(data)\
   # doctest: +IGNORE_EXCEPTION_DETAIL +NORMALIZE_WHITESPACE
     Traceback (most recent call last):
     ...
     unyt.exceptions.IterableUnitCoercionError: Received a list or
     tuple of quantities with nonuniform units:
-    [unyt_quantity(2., 'cm'), unyt_quantity(3., 'km')]
+    [unyt_quantity(2., 'cm'), unyt_quantity(3., 'g')]
     """
 
     def __init__(self, quantity_list):
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -2161,17 +2161,21 @@ def test_ones_and_zeros_like():
 
 
 def test_coerce_iterable():
-    from unyt import cm, km
+    from unyt import cm, m, g
 
     a = unyt_array([1, 2, 3], "cm")
-    b = [1 * cm, 2 * km, 3 * cm]
+    b = [1 * cm, 2 * m, 3 * cm]
+    c = [1 * g, 2 * m, 3 * cm]
 
+    assert_equal(a + b, unyt_array([2, 202, 6], "cm"))
+    assert_equal(b + a, unyt_array([2, 202, 6], "cm"))
     with pytest.raises(IterableUnitCoercionError):
-        a + b
+        a + c
     with pytest.raises(IterableUnitCoercionError):
-        b + a
+        c + a
+    assert_equal(unyt_array(b), unyt_array([1, 200, 3], "cm"))
     with pytest.raises(IterableUnitCoercionError):
-        unyt_array(b)
+        unyt_array(c)
 
 
 def test_bypass_validation():


unyt/array.py
```diff
@@ -257,7 +257,13 @@ def _coerce_iterable_units(input_object, registry=None):
         if any([isinstance(o, unyt_array) for o in input_object]):
             ff = getattr(input_object[0], "units", NULL_UNIT)
             if any([ff != getattr(_, "units", NULL_UNIT) for _ in input_object]):
-                raise IterableUnitCoercionError(input_object)
+                ret = []
+                for datum in input_object:
+                    try:
+                        ret.append(datum.in_units(ff.units))
+                    except UnitConversionError:
+                        raise IterableUnitCoercionError(str(input_object))
```
neutrinoceros, line 265: This is my one important suggestion here, see https://www.python.org/dev/peps/pep-3134/
```suggestion
                    except UnitConversionError as exc:
                        raise IterableUnitCoercionError(str(input_object)) from exc
```

unyt/exceptions.py
```diff
@@ -161,15 +161,15 @@ class IterableUnitCoercionError(Exception):
     Example
     -------
 
-    >>> from unyt import km, cm, unyt_array
-    >>> data = [2*cm, 3*km]
+    >>> from unyt import g, cm, unyt_array
+    >>> data = [2*cm, 3*g]
```
neutrinoceros, line 165: nice !

unyt/tests/test_unyt_array.py
```diff
@@ -2161,17 +2161,21 @@ def test_ones_and_zeros_like():
 
 
 def test_coerce_iterable():
-    from unyt import cm, km
+    from unyt import cm, m, g
```
neutrinoceros, line 2164:
```suggestion
    from unyt import cm, km, m, g
```

unyt/tests/test_unyt_array.py
```diff
@@ -2161,17 +2161,21 @@ def test_ones_and_zeros_like():
...
 
     a = unyt_array([1, 2, 3], "cm")
-    b = [1 * cm, 2 * km, 3 * cm]
+    b = [1 * cm, 2 * m, 3 * cm]
```
neutrinoceros, line 2167: Seems a tiny bit more fragile (in a good way) like this
```suggestion
    b = [1 * cm, 2 * m, 3 * km]
```

unyt/tests/test_unyt_array.py
```diff
@@ -2161,17 +2161,21 @@ def test_ones_and_zeros_like():
...
+    c = [1 * g, 2 * m, 3 * cm]
 
+    assert_equal(a + b, unyt_array([2, 202, 6], "cm"))
+    assert_equal(b + a, unyt_array([2, 202, 6], "cm"))
```
neutrinoceros, line 2171:
```suggestion
    assert_equal(a + b, unyt_array([2, 202, 300_003], "cm"))
    assert_equal(b + a, unyt_array([2, 202, 300_003], "cm"))
```

neutrinoceros (commented):
Here are a couple suggestions. Otherwise looks great to me !

unyt/array.py
ngoldbaum: I don’t see how this is any better. It would help in the future if you would explicitly explain why a change would help instead of suggesting that someone read a long technical document to explain why.

unyt/tests/test_unyt_array.py
ngoldbaum: Your suggested change makes zero difference to the actual thing we’re testing, no.

unyt/array.py
neutrinoceros: Sure thing. This produce a different traceback.

### with a bare `raise`
```python
try:
    1/0
except Exception:
    raise ValueError
```
one gets
``
  File "/Users/robcleme/dev/yt-project/yt/t..py", line 2, in <module>
    1/0
ZeroDivisionError: division by zero

During handling of the above exception, another exception occurred:

  File "/Users/robcleme/dev/yt-project/yt/t..py", line 4, in <module>
    raise ValueError
ValueError
```
This reads as if something unexpected happened in the except block, but it's not the case: we're raising a different error very much on purpose.

### with a `raise from`
```python
try:
    1/0
except Exception as exc:
    raise ValueError from exc
```

```
  File "/Users/robcleme/dev/yt-project/yt/t..py", line 2, in <module>
    1/0
ZeroDivisionError: division by zero

The above exception was the direct cause of the following exception:

  File "/Users/robcleme/dev/yt-project/yt/t..py", line 4, in <module>
    raise ValueError from exc
ValueError
```
This is preferable because the error message conveys the designed intention better. This is exactly the situation `raise from` was intended for

ngoldbaum: I don’t think the second version is any more helpful, sorry. It’s fine as-is.

PR #218, opened 2022-01-19
BUG: fix typos in constants

### LINKED ISSUE #205: SI prefix milli is misspelled
* unyt version: 2.8.0
* Python version: 3.7.6
* Operating System: Windows

### Description
The name of the SI prefix milli appears to be misspelled, with only a single l, as "mili".  One place (maybe the only place) where this occurs in the code is https://github.com/yt-project/unyt/blob/e86c0500e9c9f29914eed3ef7d8f11c98768eedd/unyt/_unit_lookup_table.py#L325 .

### What I Did
```
>>> import unyt
>>> unyt.milimeter
mm
>>> unyt.millimeter
AttributeError: module 'unyt' has no attribute 'millimeter'
```

neutrinoceros:
fix #205

### DIFF
--- a/unyt/_unit_lookup_table.py
+++ b/unyt/_unit_lookup_table.py
@@ -382,7 +382,8 @@ unit_prefixes = OrderedDict(
         ("da", (1e1, "deca")),
         ("d", (1e-1, "deci")),
         ("c", (1e-2, "centi")),
-        ("m", (1e-3, "mili")),
+        ("m", (1e-3, "mili")),  # typo, kept for backward compatibility
+        ("m", (1e-3, "milli")),
         ("µ", (1e-6, "micro")),  # ('MICRO SIGN' U+00B5)
         ("u", (1e-6, "micro")),
         ("μ", (1e-6, "micro")),  # ('GREEK SMALL LETTER MU' U+03BC)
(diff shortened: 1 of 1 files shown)


neutrinoceros:
also fixes #222
Switching to draft while I try to think about a nicer way to fix these without breaking existing user code. Of course a simple way to do that would be to simply have both the correct and the historic-and-mispelled versions coexist for a while. The hard part is making the erroneous version visibly deprecated.

neutrinoceros:
also rebased on top of #219 to stabilise CI

neutrinoceros:
I seems tricky to properly deprecate for these, so I've switched the PR to a more conservative and backward compatible approach for now

unyt/tests/test_unyt_array.py
```diff
@@ -828,6 +832,23 @@ def test_ytarray_pickle():
         assert_equal(float(data.units.base_value), float(loaded_data.units.base_value))
 
 
+@pytest.mark.xfail(
```
jzuhone, line 835: @neutrinoceros is this related to the rest of the PR?

jzuhone: Never mind, I saw you had to rebase on #219. We'll get this sorted soon.

jzuhone (approved):

Issue #193, opened 2021-08-07
Use pre-commit hooks to apply linting rules


matthewfeickert:
### Description

As noted in https://github.com/yt-project/unyt/pull/191#issuecomment-894608515 there are some PR that fail linter tests. As most of these linter tests just require a human to manually run a tool, like `flake8` or `black`, it seems like it would be helpful to use `pre-commit` hooks to run this for the user.

In my own projects what I've found super useful for keeping pre-commit hooks up to date and for saving time is to enable [pre-commit.ci](https://pre-commit.ci/) which will allow for pushes back to a PR if pre-commit hooks fail with attempts at resolving those failures.

There are some [examples over on the Scikit-HEP website](https://scikit-hep.org/developer/style#pre-commit) (in addition to all the examples the pre-commit team has).

neutrinoceros:
We use pre-commit.ci for yt itself now, I think it is worth migrating to it for unyt to, in particular because it saves the trouble of syncing installed versions for linters for all devs.

neutrinoceros:
CI is currently broken because black 22.1 made some intentional style changes that are not currently applied to unyt.
I'm going to pin it to the previous version in tex.ini for now but this is not a viable solution long term. Using `pre-commit.ci` would be much more viable.

PR #219, opened 2022-02-22
TST: use sympy pre-releases in bleeding-edge workflow

neutrinoceros:
Looks like sympy 1.10rc2 breaks unyt, see https://github.com/sympy/sympy/issues/23148

I propose we start using sympy pre-releases in the daily "bleeding-edge" workflow so these problems are always caught in a timely fashion.


oscarbenjamin:
I intend to make more frequent pre-releases of sympy in future so this would be useful to pick up on issues long before the main release.

neutrinoceros:
Looks like normal CI is broken too, for other reasons, I don't know if I'll be able to investigate this today.

neutrinoceros:
So one of the problems was that black made some style changes since the last time CI was run successfully. This problem and a possible solutions are discussed in #193. Pinning black for now.
Another problem is that the last version of pytest-doctestplus that is available for Python 3.6 is broken by the latest version of pytest
https://github.com/astropy/pytest-doctestplus/issues/170

I'll pin that too.

neutrinoceros:
I didn't manage to constrain pytest<7 just for Python 3.6 so I'm pinning it globally in tox.ini

neutrinoceros:
Ran the workflow on my fork again with sympy 1.10rc3 and it passed https://github.com/neutrinoceros/unyt/runs/5345360352?check_suite_focus=true

.github/workflows/ci.yml
```diff
@@ -5,6 +5,8 @@ on:
     branches:
     - master
   pull_request:
+  schedule:
+    - cron:  '1 0 * * *'
```
chrishavlin, line 9: What's the benefit to running the standard ci every day in addition to the beleeding-edge?

chrishavlin (approved):
Thanks for figuring this out!

.github/workflows/ci.yml
neutrinoceros: The bleeding edge workflow didn't detect the formatting changes from black, so it's clearly not covering everything that normal CI does. My reasoning is that it's easier to just run both daily than trying to cover everything with bleeding edge.

neutrinoceros:
To @oscarbenjamin's request (https://github.com/sympy/sympy/issues/22241#issuecomment-1061302909), I'm adding a test for unpicking unyt arrays with different versions of sympy. Hopefully this should help remove some of the maintenance burden from sympy.

oscarbenjamin:
> I'm adding a test for unpicking unyt arrays with different versions of sympy.

The best thing would be to add a test along these lines to sympy itself rather than unyt. I don't know what's in these pickles but can you make a representative example that doesn't actually require unyt to be installed?

Ideally I'd want a test that's just like
```python
pickle = b'...'
assert loads(pickle) == ...
```
Then we could add pickles for the same expression generated with all previous versions of sympy and see that they all unpickle correctly on current sympy master.

neutrinoceros:
I may give it a try in a few days. ATM I'm not sure how to minimise the test case without missing the point as we did 6 month ago.

neutrinoceros:
@jzuhone could you please have a look ? CI is still very unstable on unyt without these patches

jzuhone (approved):

Issue #226, opened 2022-04-21
Unyt is using a truncated version of Avogadro's number?


bc118:
It seems like Unyt is using a truncated version of Avogadro's number, being 6.0221e23.  Is this the case?  If so, I recommend the following.

The seemingly more accurate Avogadro's number is 602,214,141,070,409,084,099,072 and located here (https://www.americanscientist.org/article/an-exact-value-for-avogadros-number). 

I recommend double-checking and adding the most accurate value of Avogadro's number above if this is indeed the issue.

neutrinoceros:
Here's where the constant is hardcoded. 
https://github.com/yt-project/unyt/blob/ec5b3301c110787c9a67c600a66d8051ab8c831e/unyt/_physical_ratios.py#L103
As you can see, it has a couple more digits than you were guessing. It's also stored as a `float`, which limits the precision with which we can write it
we could change it from
`6.0221408577399995e+23` (effective current value)
to
`6.022141410704091e+23`
though this may qualify as a breaking change for other users. Personally I would agree to update it anyway, since it's technically a fix.

neutrinoceros:
The downstream compatibility issue is akin to https://github.com/yt-project/unyt/issues/144
However I don't think that `avogadros_number` is very likely to be used in yt.

bc118:
Thank you for your quick response and information.  I added by additional thought below.

The chemistry, chemical engineers, and molecular simulations people use Avogadro's number a lot as it is key to chemistry, so we have noticed the slight difference.  Therefore, we believe it is vital to use the correct number.

When checking the accuracy of our Unyt calculations compared to manual checks, the manual constants must be modified to account for the error in the Unyt Avogadro's number, or other measures are taken (for example, using "isclose" to make the comparison).

At a minimum, I believe this error in Avogadro's number should be listed in the Uynt documents because it is likely to be noticed by others and will help them quickly find the discrepancy.

neutrinoceros:
I understand. Do you use the constant directly or is it used under the hood of some application (unyt itself ?) that you have no control over ?
In the former case, we could also ship the more accurate value under a different name (and of course document it).

Whatever the situation, I'll make sure this matter is discussed and that we have a solution for the next release.

bc118:
We use it in both cases.  Users can use Unyt and put the values into our software, and/or under the hood of the software that we write and distribute.

bc118:
[...]

PR #228, opened 2022-06-08
various additions to constants and units

jzuhone:
This PR:

* Adds the dimension `angular_frequency` and the unit `rpm`.
* Adds the [Rydberg constant and Rydberg unit](https://en.wikipedia.org/wiki/Rydberg_constant).
* Removes the hard-coded value of the `elementary_charge` in two places and puts it in a variable in `physical_ratios`.
* Updates the value of Avogadro's number (closes Issue #226)

### DIFF
--- a/unyt/_physical_ratios.py
+++ b/unyt/_physical_ratios.py
@@ -101,8 +101,13 @@ elementary_charge = 1.6021766208e-19
 mu_0 = 4.0e-7 * np.pi
 # permittivity of Free Space
 eps_0 = 1.0 / (speed_of_light_m_per_s ** 2 * mu_0)
-avogadros_number = 6.022141410704091e+23
-rydberg_constant = 0.125*mass_electron_kg*elementary_charge**4/(eps_0**2*planck_mks**3*speed_of_light_m_per_s)
+avogadros_number = 6.022141410704091e23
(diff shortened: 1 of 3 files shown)


unyt/_physical_ratios.py
```diff
@@ -115,6 +122,7 @@
 kelvin_per_rankine = 5.0 / 9.0
 watt_per_horsepower = 745.69987158227022
 erg_per_s_per_watt = 1e7
+rydberg_unit = planck_mks * speed_of_light_m_per_s * rydberg_constant
```
neutrinoceros, line 125: same here ?
```suggestion
rydberg_unit_mks = planck_mks * speed_of_light_m_per_s * rydberg_constant
```
```suggestion
rydberg_unit = planck_mks * speed_of_light_m_per_s * rydberg_constant
```

neutrinoceros (commented):
Shouldn't the names of each new float constant suggest its unit or unit system as in other similar constants ?

Seems to me the following substitutions would be appropriate
```
elementary_charge -> elementary_charge_C
rydberg_constant -> rydberg_constant_mks
rydberg_unit -> rydberg_unit_mks
```

neutrinoceros (approved):

Issue #230, opened 2022-06-10
`np.divide.reduce` gives incorrect units


kyleaoman:
* unyt version: v2.8.0+82.geb511a2 (master commit eb511a2a2455e1d2d2fbccdbfbff7b142cb5625f)
* Python version: 3.10.0
* Operating System: Linux Mint 20.3 (5.4.0-113-generic)

### Description

Stumbled on an incorrect implementation while working on a subclass of unyt. The `unyt_array.__array_ufunc__` method incorrectly implements the calculation of units for the `np.divide.reduce` method. Given an array [a, b, c] with units u, the method gives a / b / c and therefore the units should be u / u / u = u ** -1, not (u**-1) ** 3 = u ** -3 as currently implemented. In other words, it's not symmetric with `np.multiply.reduce`, where u ** 3 is correct.

### What I Did

```
>>> import numpy as np
>>> import unyt
>>> ua = unyt.unyt_array([2., 2.], unyt.kpc)
>>> print(np.divide.reduce(ua))
1.0 kpc**(-2)
>>> # expected dimensionless, or in general kpc^(2-N) where N is length of array, instead of kpc^(-N)
```

PR #231, opened 2022-06-10
Bugfix for `np.divide.reduce`

kyleaoman:
Simple fix to address #230. Replaced `POWER_SIGN_MAPPING` with `POWER_MAPPING` and had this return simple functions (`lambda`s) to compute the exponent instead of just giving the sign of the exponent, which was used in the previous incorrect implementation. Added some unit tests for `np.divide.reduce` (and `np.multiply.reduce` for good measure), these fail before changes and pass afterwards. Also ran `black` and committed the result before starting so I could start with a "clean" set of tests. The functional changes are in the other two commits.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -139,7 +139,7 @@ from unyt.unit_registry import (
 )
 
 NULL_UNIT = Unit()
-POWER_SIGN_MAPPING = {multiply: 1, divide: -1}
+POWER_MAPPING = {multiply: lambda x: x, divide: lambda x: 2 - x}
 
 __doctest_requires__ = {
     ("unyt_array.from_pint", "unyt_array.to_pint"): ["pint"],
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Hey ! thank you for your contribution (and the report).
Can you please only checkin the bits that are actually relevant to the fix ? Right now it seems you packed a bunch of unrelated style fixes, my guess is that you are using a more recent version of black than the one we have pinned here.

neutrinoceros:
For completeness, we use `black==21.12b0` at the moment

kyleaoman:
Dropped that commit.

neutrinoceros (approved):
Looks great ! thank you for adding tests and finding a minimal and elegant fix.
I'll defer to @jzuhone to merge this

jzuhone (approved):

PR #239, opened 2022-07-08
define_unit override argument

rjfarber:
I'm not sure this is fully desirable (or maybe this capability exists and I'm just missing it) but I'm currently in the situation where I messed up a custom defined unit and wanted to simply override the definition (similar to how derived_fields can be redefined).

So I added an optional argument in `def define_unit(...,override=False)` and don't trigger the raise if override is True. I have to run to a meeting but thought I might as well kick off this discussion rather than spend a lot of time on making this a complete PR right now in case this is not desirable.

Anyway happy Friday!

### DIFF
--- a/unyt/unit_object.py
+++ b/unyt/unit_object.py
@@ -1038,7 +1038,7 @@ def _validate_dimensions(dimensions):
 
 
 def define_unit(
-    symbol, value, tex_repr=None, offset=None, prefixable=False, registry=None
+    symbol, value, tex_repr=None, offset=None, prefixable=False, registry=None, override=False
 ):
     """
     Define a new unit and add it to the specified unit registry.
@@ -1064,6 +1064,9 @@ def define_unit(
         unit object will be added as an attribute to the top-level :mod:`unyt`
         namespace to ease working with the newly defined unit. See the example
         below.
+    override : boolean, optional
+        Whether or not to allow an override to an existing unit in the given 
+        unit registry.
 
     Examples
     --------
@@ -1080,7 +1083,7 @@ def define_unit(
 
     if registry is None:
         registry = default_unit_registry
-    if symbol in registry:
+    if symbol in registry and not override:
         raise RuntimeError(
             "Unit symbol '%s' already exists in the provided " "registry" % symbol
         )


neutrinoceros (commented):
Hi @rjfarber 
This looks like a reasonable request to me and your implementation is sound. I'll take a deeper look tomorrow to make sure this is the simplest way to do it (but it seems likely).
For now I just have a couple minor suggestions:
- `force_override` or `allow_override` would make slightly more sense to me as an argument name, because `override=True` reads as something that's never a no-op.
- I'd like that any new arguments was keyword only, because it's much easier to maintain for us in the future.

neutrinoceros:
So I don't think there's any documented/supported way to override pre-defined units at the moment.
Within my admittedly limited understanding of the code involved, I also cannot foresee any fundamental problem with this approach.
Here's what I could grasp: under the hood, this function updates a `UnitRegistry` instance, which is a relatively flat data structure: I don't think that it preserves a hierarchy of dependencies between units. My conclusion is that there should be no risk of getting stuck with a corrupt state with cyclic references and whatnot.

If you want to pursue this, feel free to. Most importantly we would need some documentation and tests.

jzuhone:
So the `UnitRegistry` class has a `modify` method:

https://github.com/yt-project/unyt/blob/10ab887196fe490388fdf4f7eafb3853160f44b6/unyt/unit_registry.py#L199-L229

I'm not sure if this does what you want it to or not.

neutrinoceros:
Oh I missed that, thanks for pointing it out !
So `UnitRegistry.modify` only works for previously defined units, while `define_unit` currently only works for adding new units. I think making `define_unit` the go-to function to do both is a good idea and solidifies the API.

`unyt.define_unit` is a top level user-facing function which encapsulates lower level concepts such as `UnitRegistry` management, it may make sense to plug a call to `UnitRegistry.modify` in the override implementation.

jzuhone:
I guess I have a question about the use case--@rjfarber, if you messed up the unit, is there a reason we don't just fix the unit in the script and then re-run? Or is there a reason we should start with an incorrect value and then change it?

I'm only a bit hesitant to introduce this functionality because in normal cases we actually don't want to redefine units--a kpc is going to be a kpc, and if the standard changes then we change the hard-coded value instead. `UnitRegistry.modify` is there mainly for code units. I worry making it too easy to change units could lead to unpredictable behavior.

However, maybe I am not understanding the use case well enough.

rjfarber:
Hi John,

Regarding use case, if you're running in a jupyter notebook and already
have a decent amount of data loaded (can take minutes to load) it's rather
annoying to shutdown the script, reload the data, etc. Really slows down
the development process. And I ran into this for dealing with
non-dimensional units (cloud radii, cloud mass etc) - which is a problem
for dimensionalized codes like FLASH as much as for code-unit codes like
Athena.

It does sound like UnitRegistry.modify can handle this case, but I agree
with Clément that I think it'd be nicer to have it all under define_unit.
Although I'll definitely to the judgement of you two and the other
long-time yt devs here :)

Regarding making it too easy to change units, well if the user decides to
use a "force_override" method hopefully they know what they're doing.
Personally I'm a fan of giving the user absolute power to `sudo rm -rf /`
vs. coddling the user, but that gets philosophical I'll admit :)

Also sorry everyone for being MIA! I'll plan to return to this tomorrow :)

Best,
--------
Ryan

neutrinoceros:
Furthermore giving that power to users is perfectly in line with how `yt.derived_field` works. One could argue that yt and unyt don't have to follow the exact same design principles but they do share a significant fraction of their users ...

jzuhone:
@rjfarber I haven't forgotten about this, sorry for the delay. I think in a couple of weeks we should try getting it merged in

rjfarber:
Okay cool! Did you and Clément come to a decision/understanding about a
"allow_override" argument vs. modify_unit new function regarding
implementation?

Best,
--------
Ryan

neutrinoceros:
We haven't discussed it in other channels. I'm still inclined to think an `override`-like argument to an existing function is the best option.

jzuhone:
@rjfarber so I've come around to this idea--but I would like to see a couple of things (potentially) added here:

1. It would be nice if we could indicate that a unit is being overridden somehow. But we don't have a logger, and I am hesitant to add one, especially since we don't have more use cases. And we shouldn't use a print statement or issue a warning either. So not sure if there is a good way to log this, open to suggestions. 
2. We just need a short bit in the docs about this, one or two sentences is probably fine. If you need help with that let me know.

neutrinoceros:
> we don't have a logger, and I am hesitant to add one, especially since we don't have more use cases. And we shouldn't use a print statement or issue a warning either. So not sure if there is a good way to log this, open to suggestions.

Agreed that adding a logger for this seems like a long shot, and a plain print statement would be terrible style, but what's wrong with a warning ? If you fear this would be too noisy, I can see two ways we could add a warning and keep it "quiet" in most cases:

- we could subclass (or use) builtin `ResourceWarning`, which aren't showed by default but will be treated as errors in CI here and downstream, greatly reducing the risk that this ends up in production code by accident https://docs.python.org/3/library/exceptions.html#ResourceWarning
- or we could add _another_ option to silence the warning. I must say I'm not a fan of this option.

rjfarber:
Yup, I'll definitely add to the docs!

Regarding adding some indication to the user that an override is taking
place, I agree a print statement would not be ideal. I'm cool with
Clément's suggestion of using a warning. I guess you both will be horrified
to hear this, but one my favorite stackoverflow answers is:

I don't condone it, but you could just *suppress all warnings* with this:

import warnings
warnings.filterwarnings("ignore")

(https://stackoverflow.com/questions/14463277/how-to-disable-python-warnings
)

Best,
--------
Ryan

neutrinoceros:
>I guess you both will be horrified
to hear this, but one my favorite stackoverflow answers is:
>
>I don't condone it, but you could just *suppress all warnings* with this:

First of all: 😱 
Second of all: we're well aware that's a thing, and we'd prefer to not put our users in a position where they would need or want this.

rjfarber:
Don't worry I mostly do that because of matplotlib I think 😅

Anyway, I made modifications to the define_unit docstring and now I
understand John's concern better I think - since I guess the define_unit
was intended as creating brand new units rather than modifying them, so I
had to change the wording kind of repetitively.

I thought this modification had worked in my original use case, but I'm not
sure where that script is. And when trying to write a non-Jupyter example
(to add to the examples in the docstring), it doesn't seem like my unit
actually gets updated 🤔

Although also the tests/define_unit_test.py is failing now on the "Test
custom registry" part.
Just thought I'd update you all that I'm back at this! Hopefully I can
figure the above problems out with fresh eyes tomorrow :)

Best,
--------
Ryan

PR #240, opened 2022-07-09
ENH: Give the option to change the value of the solar metallicity

jzuhone:
This PR implements a function that allows one to change the value of the solar metallicity conversion factor. The default will remain the same, consistent with cloudy. 

Forthcoming:

- [ ] docs
- [ ] tests


jzuhone:
I stumbled over something when creating this that opens up a bunch of complications (not a bug), so I'm going to make this a draft for now.

jzuhone:
I'm closing this in favor of a less invasive idea, which I'll propose later.

PR #244, opened 2022-07-12
STY: flynt + pyupgrade

neutrinoceros:
Additional (and final) style alignment with yt.
This is low priority, I just add a couple minutes to kill


ngoldbaum:
Sure why not

PR #245, opened 2022-07-12
TST: unpin pytest

neutrinoceros:
Cleanup. According to the comment I left there, this pin isn't necessary now that support for Python 3.6 was dropped.


neutrinoceros:
refreshing CI to make sure this is still relevant

Issue #203, opened 2021-10-06
Zero Exponent in unyt


JBorrow:
* unyt version: unyt v2.8.0
* Python version: Python 3.9.7
* Operating System: MacOS

### Description

Taking unyt_quantity to zero exponent returns original units and not (dimensionless).

Is this expected behaviour?

### What I Did

```
import platform
import unyt

print("Python", platform.python_version())
print("unyt", unyt.__version__)

base_quantity = unyt.unyt_quantity(112.0, "g")

print("Raw", base_quantity, "expected g")
print("Small Exponent", base_quantity**0.03, "expected g**0.03",)
print("Zero Exponent (float)", base_quantity**(0.0), "expected (dimensionless)")
print("Zero Exponent (int)", base_quantity**0, "expected (dimensionless)")
```

Output:
```
Python 3.9.7
unyt v2.8.0
Raw 112.0 g expected g
Small Exponent 1.1520638270207721 g**(3/100) expected g**0.03
Zero Exponent (float) 1.0 g expected (dimensionless)
Zero Exponent (int) 1.0 g expected (dimensionless)
```

ngoldbaum:
I think you’ll need to edit the `__pow__` implementation for the `Unit` class: https://github.com/yt-project/unyt/blob/e86c0500e9c9f29914eed3ef7d8f11c98768eedd/unyt/unit_object.py#L469

You may also need to touch this code path in the `unyt_array` class: https://github.com/yt-project/unyt/blob/e86c0500e9c9f29914eed3ef7d8f11c98768eedd/unyt/array.py#L1758

I definitely agree this is a bug. I don’t have time right now to work on a fix.

JBorrow:
Looks like when using a power of zero, we default to the ufunc `_ones_like`. I'm guessing this is a numpy optimisation, and we'll need to just make sure that the right units (dimensionless) gets associated in this case.

```
In [13]: x = unyt.unyt_quantity(4.0, "g")

In [14]: pdb.run('x**0.0')
> <string>(1)<module>()
(Pdb) s
--Call--
> /Users/jborrow/Documents/OpenSource/unyt/unyt/array.py(1694)__array_ufunc__()
-> def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
(Pdb) ufunc
<ufunc '_ones_like'>
```

ngoldbaum:
Ah, fun! That might make things harder if there’s
no way to determine if the original ufunc was a pow.

[...]

PR #249, opened 2022-07-15
BUG: fix inheritance in unyt_array.__pow__

neutrinoceros:
This change is a solution to a downstream issue in yt:
https://github.com/yt-project/yt/issues/4023

it's a direct follow up to #204 

I'm not 100% sure it's the *right* solution, as I don't understand the original implementation using `__array_func__`, so maybe I'm just missing something, in which case the problem should be resolved on the yt side instead.

@JBorrow I know it's been a little while, any chance you could explain your initial implementation please ?

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1713,10 +1713,9 @@ class unyt_array(np.ndarray):
         if p == 0.0:
             ret = self.ua
             ret.units = Unit("dimensionless")
+            return ret
         else:
-            ret = self.__array_ufunc__(power, "__call__", self, p)
-
-        return ret
+            return super().__pow__(p)
 
     #
     # Start operation methods
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1702,7 +1702,7 @@ class unyt_array(np.ndarray):
             pass
         return ret
 
-    def __pow__(self, p):
+    def __pow__(self, p, mod=None, /):
         """
         Power function, over-rides the ufunc as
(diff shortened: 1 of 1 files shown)


neutrinoceros:
So I digged a little further and I think I get why it was implemented this way: given `arr` a `unyt_array` instance, as of unyt 2.8, the `__pow__` protocol was only called for `are**x` *if* `x==0`, and it's the only case that was problematic (#203). However the situation changes if `unyt_array` overrides the base `__pow__` protocol from `np.ndarray`, and now `__pow__` gets called for every value of `x`. @JBorrow very likely understood this and chose an implementation that made sense to preserve backward compatibility, redirecting to `self.__array_ufunc__` for `x != 0`.

Subclassing `np.ndarray` proves however more complicated than expected, and this did not _actually_ preserve backward compatibility as seen in yt https://github.com/yt-project/yt/issues/4023
My corrected implem leaves all responsabilities to numpy and works better in that sense.

While I was researching this problem, I also realised that the `__pow__` method normally takes an optional `mod` argument https://numpy.org/doc/stable/reference/generated/numpy.ndarray.__pow__.html?highlight=__pow__#numpy.ndarray.__pow__
so I "fixed" this too (using quotation marks here because I couldn't find a single working example using this argument, even with pure `np.ndarray`s)

neutrinoceros:
Thank you very much !
@jzuhone I think this is now ready :)

jzuhone (approved):

PR #250, opened 2022-07-16
ENH: implement lazy imports for optional dependencies

### LINKED ISSUE #27: Optimize importing unyt

neutrinoceros:
This is another step to resolve #27
Overall, this gets us to a state where core dependencies (numpy + sympy) are responsible for at least 50-60 % of unyt's import time.

This refactor is similar to my work in yt https://github.com/yt-project/yt/pull/3935
It allows for optional dependencies to never be imported before they are actually requested by users, reducing unyt's own startup time when heavy optional dependencies are installed.

Making matplotlib a true "on-demand" import took a little more work, but now `import unyt` doesn't cause `matplotlib` to be imported too when available, reducing the startup time by about 15% to 25% with respect to unyt 2.9.2 (on my machine).

### DIFF
--- a/unyt/_on_demand_imports.py
+++ b/unyt/_on_demand_imports.py
@@ -10,6 +10,10 @@ A set of convenient on-demand imports
 # The full license is in the LICENSE file, distributed with this software.
 # -----------------------------------------------------------------------------
 
+from functools import wraps
+from importlib.util import find_spec
+from typing import Type
+
 
 class NotAModule:
     """
@@ -21,8 +25,7 @@ class NotAModule:
     def __init__(self, pkg_name):
         self.pkg_name = pkg_name
         self.error = ImportError(
-            "This functionality requires the %s "
-            "package to be installed." % self.pkg_name
+            f"This functionality requires the {self.pkg_name} package to be installed."
         )
 
     def __getattr__(self, item):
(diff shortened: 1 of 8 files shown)


neutrinoceros:
this now includes changes for dask, recently added with #185 (ping @chrishavlin)

neutrinoceros:
this is now based on #272

neutrinoceros:
@jzuhone @chrishavlin can I request your reviews on this ?

tox.ini
```diff
@@ -57,9 +57,9 @@ deps =
     dask[array,diagnostics]>=2021.04.1
 depends = begin
 commands =
-    # don't do doctests in rst files due to lack of way to specify optional
```
jzuhone, line 60: Is there a reason you took this out here? It wasn't mentioned in the description and it seems unrelated to load times.

jzuhone (approved):
Curious about the doctests bit, but otherwise looks good.

tox.ini
neutrinoceros: I can't recall why I did that. Let's try to revert this part and see if anything fails.

neutrinoceros: welp, nothing breaks. I'll just clean up the history

chrishavlin (approved):
Looks good! Left a couple minor comments.

One additional question -- why'd you put the `unyt_arrayConverter` class in `_mpl_array_converter.__init__.py` instead of a new module, `_mpl_array_converter.py` ?

unyt/_on_demand_imports.py
```diff
@@ -31,182 +34,142 @@ def __getattr__(self, item):
...
-    _log = None
-    _units = None
-    _version = None
+    def __init_subclass__(cls):
+        if not cls.__name__.endswith("_imports"):
+            raise TypeError(f"class {cls}'s name needs to be suffixed '_imports'")
+
+    def __new__(cls):
+        if cls is OnDemand:
+            raise TypeError("The OnDemand base class cannot be instanciated.")
```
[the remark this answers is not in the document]
neutrinoceros: good catch !

unyt/_on_demand_imports.py
```diff
@@ -31,182 +34,142 @@ def __getattr__(self, item):
...
-                from matplotlib import pyplot
-            except ImportError:
-                pyplot = NotAModule(self._name)
-            self._pyplot = pyplot
-        return self._pyplot
+        from matplotlib import pyplot
+
+        return pyplot
 
     @property
```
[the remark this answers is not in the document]
neutrinoceros: not intentional, thanks ! (same with `use`)

neutrinoceros:
> One additional question -- why'd you put the unyt_arrayConverter class in _mpl_array_converter.__init__.py instead of a new module, _mpl_array_converter.py ?

I can't remember. I would guess I went through several iterations and that form was the first one with which I managed to pass tests, for some reason. Now I agree that a module is better than a mono-module package, and I've checked that it does not in fact break anything, so let's roll with that.

PR #251, opened 2022-07-16
ENH: reduce import time by skipping version checking

neutrinoceros:
This is a simple optimization for unyt's import time (address #27)

I noticed that importing `pkg_ressources` inside a try-block costed about 10% of the current import time, which seems unreasonable for the marginal improvement it brings to error message clarity.

### DIFF
--- a/unyt/__init__.py
+++ b/unyt/__init__.py
@@ -43,33 +43,6 @@ top-level ``unyt`` namespace:
 # The full license is in the LICENSE file, distributed with this software.
 # -----------------------------------------------------------------------------
 
-try:
-    import numpy as np
-
-    try:
-        from pkg_resources import parse_version
-
-        npv = np.__version__
-        if parse_version(npv) < parse_version("1.13.0"):  # pragma: no cover
-            raise RuntimeError(
-                "The unyt package requires NumPy 1.13 or newer but NumPy %s "
-                "is installed" % npv
-            )
-        del parse_version, npv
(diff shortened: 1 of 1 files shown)


munkm:
Is there an alternative user-friendly import error now that this has been deleted? In my experience, lots of users get very confused by errors related to package imports when they don't have clear messages.

neutrinoceros:
Not without impacting the import time for everyone, or making unyt harder to maintain (something with release branches and a less optimized dev branch), I think.

munkm:
Yeah I agree that going with two dev branches is not a good idea. I just worry that this negative impact to users isn't great.

neutrinoceros:
For transparency, I genuinely think those error messages were completely impossible to trigger if you installed unyt from pypi or conda forge, unless you willingly uninstalled sympy or numpy after that. Happy to be proven wrong and reconsider.
Also, for full discloser, only the `pkg_resources` import had an actual impact on import time, we probably *could* reintroduce import error wrapping for numpy and sympy with at no noticeable cost.

jzuhone:
@munkm so in this particular case I looked over the PR and agreed that it would be very difficult for these errors to be triggered unless you were force-removing sympy or numpy. And numpy < 1.13 is very old now. Even for a typical source installation, this would not be likely. In general I agree that we need to think about this kind of thing, however.

munkm:
Cool. Thanks both!

PR #252, opened 2022-07-16
ENH: reduce redundant definitions for aliased unit symbols

neutrinoceros:
Shaves an additional 5% off of unyt's import time (address #27).
However this might not be worth it because it's not strictly backward compatible: it makes aliased unit symbols actually point to a single object instead of several identical objects, and maybe that's undesirable somehow ?
Anyway I'll still submit this proposal since it doesn't seem to break any existing test.


neutrinoceros:
it's not clear that this actually saves anything measurable when combined with #251 and #252 so I'll keep this one as a draft for now

neutrinoceros:
spent some more time on this and I now think it may never be worth the effort, as the change in perf is far bellow noticeable levels. It seems likely that upstream changes from Python and sympy would be far more efficient.

PR #253, opened 2022-07-16
ENH: don't import from top level sympy namespace

neutrinoceros:
Sympy is a major contributor to unyt's import time.
By importing sympy elements from their respective modules, we should in principle be able to avoid importing the whole sympy namespace and reduce unyt's import time.
Currently this has no actual impact because of a seemingly circular import in sympy:

```python
from sympy.core.sympify import sympify
```
imports the whole `sympy` module as a side effect.
We also cannot nest these specific imports in functions because `sympify` is used at unyt's top level to define some important objects.
So, maybe there's room to improve the situation upstream. In the mean time, it doesn't hurt to make these imports as specific as possible here.

### DIFF
--- a/unyt/_parsing.py
+++ b/unyt/_parsing.py
@@ -16,7 +16,10 @@ parsing utilities
 
 import token
 
-from sympy import Basic, Float, Integer, Rational, Symbol, sqrt
+from sympy.core.basic import Basic
+from sympy.core.numbers import Float, Integer, Rational
+from sympy.core.symbol import Symbol
+from sympy.functions.elementary.miscellaneous import sqrt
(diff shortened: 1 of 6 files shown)


neutrinoceros:
I reported the upstream problem at https://github.com/sympy/sympy/issues/23787

asmeurer:
We really would prefer for upstream users to use the top-level namespace rather than specific submodules. It shouldn't really matter for performance, because every submodule will import all of SymPy.

neutrinoceros:
thank you for following me back here ! :)
I guess there's no point in hanging to this patch then

PR #255, opened 2022-07-17
ENH: migrate from versioneer to setuptools_scm

### LINKED ISSUE #254: Migrate away from versioneer ?
### Description

I found that the following line contributes significantly to unyt's import time
https://github.com/yt-project/unyt/blob/10ab887196fe490388fdf4f7eafb3853160f44b6/unyt/__init__.py#L122

using tuna, we can vizualise the cost of this line: it represents essentially 100% of the overhead from `unyt/__init__.py` (blank space on the right) 
Screenshot 2022-07-17 at 13 08 36

so it's about as much as numpy (about a couple 0.1s), which seems very expensive, especially for releases, where the result of this function call is predictable and immutable.

It seems a good solution to this problem would be to migrate to `setuptools_scm` for the specific task of compiling version tags.
The key difference is that `setuptools_scm` performs the operation at build time instead of runtime.
The costs of doing this migration would be:
- obviously some time investment to actually perform the migration (I don't mind doing it)
- the `unyt.__version__` string in dev copies would only be updated upon recompilation, as opposed to being always correct at runtime as is the case now.

So there's a little cost in development comfort, but I think it's justified by the gain on the user side.

Related to, but more specific than #27

neutrinoceros:
closes #254


neutrinoceros:
Turns out my initial motivation for doing this doesn't hold; the import time penalty from versioneer doesn't apply to releases. If there's no real value in this change I wouldn't mind seeing it closed alongside the issue.

neutrinoceros:
In retrospect, it seems like the actual gains from this are
- a noticeable startup time optimisation _for dev versions_. This still has some value IMO because it make other, user-affecting startup problems easier to spot and debug.
- possibly less maintenance weight (unclear)

neutrinoceros:
I'm not going to hold off this PR now that I know it doesn't improve the situation for users.

PR #277, opened 2022-09-29
restore ability to run tests when dask isn't installed

ngoldbaum:
Right now if you try to run the tests with `pytest --doctest-plus --doctest-modules` when dask isn't installed you will get an error:

```
_____________________ ERROR collecting unyt/dask_array.py ______________________
unyt/dask_array.py:20: in <module>
    _dask_Array = dask.array.core.Array
unyt/_on_demand_imports.py:32: in __getattr__
    raise self.error
E   ImportError: This functionality requires the dask package to be installed.
```

This is happening because pytest has discovered doctests in `unyt/dask_array.py` but that module imports stuff from dask at top-level so the test discovery is failing. To fix that we need to make those top-level imports have a fallback when dask isn't installed. We also need to tell pytest to skip the doctests in that file. There was code already to handle that, but it wasn't using the scoping that the `__doctest_requires__` feature of `pytest-doctestplus` requires correctly. It also didn't include all the doctests that needed to be skipped.

All of this wasn't being caught because `dask` was being installed for all of the `tox` jobs, including the one that tests a minimal set of dependencies. To fix the tests going forward I removed `dask` from that `tox` job. I also removed the unnecessary minimum version constraints on the other `tox` jobs and made it so that the job that tests the earliest supported versions of dependencies to use an exact version constraint.

### DIFF
--- a/unyt/dask_array.py
+++ b/unyt/dask_array.py
@@ -13,12 +13,15 @@ import unyt.array as ua
 from unyt._on_demand_imports import _dask as dask
 
 __doctest_requires__ = {
-    ("dask_array.unyt_from_dask"): ["dask"],
+    ("unyt_from_dask", "reduce_with_units", "unyt_dask_array.to_dask"): ["dask"],
 }
 
 
-_dask_Array = dask.array.core.Array
-_dask_finalize = dask.array.core.finalize
+if dask.__is_available__:
+    _dask_Array = dask.array.core.Array
+    _dask_finalize = dask.array.core.finalize
+else:
+    _dask_Array, _dask_finalize = object, None
 
 # the following attributes hang off of dask.array.core.Array and do not modify units
 _use_unary_decorator = {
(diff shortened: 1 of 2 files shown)


unyt/dask_array.py
```diff
@@ -13,12 +13,15 @@
 from unyt._on_demand_imports import _dask as dask
 
 __doctest_requires__ = {
-    ("dask_array.unyt_from_dask"): ["dask"],
+    ("unyt_from_dask", "reduce_with_units", "unyt_dask_array.to_dask"): ["dask"],
```
chrishavlin, line 16: Ah, ya. This was totally wrong! I suspect I got it working early on and then forgot to update it as the PR evolved. And then missed it cause of adding dask to all the tox envs as you described. Thanks for the fix!

chrishavlin (approved):
Thanks for the fixes!

PR #278, opened 2022-10-01
make unyt_quantity.from_string parse ints

ngoldbaum:
Sorry for getting this out on a Friday night, no need to take a look at this immediately.

I noticed that `unyt_quantity.to_string` wasn't being tested, so I modified the `from_string` test to round-trip back to a string via `to_string`. In the process, I noticed that ints weren't being loaded as I expected them to. I think it's probably closer to what a user would naively expect if we don't convert everything to float.

Since this behavior was included in a release it's possible some downstream users are relying on it. That said, I don't immediately see how making this change would cause much breakage since numpy and unyt can mix ints and floats with no issues.

Finally, the regex originally used parses `1.` as `1`, not `1.0` (python parses `1.` as `1.0`). Until making this change that didn't matter since the parsed number would be cast to a float anyway. I fixed the parsing issue by using a different scientific notation regex. The old one could probably be fixed but I couldn't figure that out, happy to do it that way if someone who is better at regex squinting can suggest it.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -149,7 +149,9 @@ __doctest_requires__ = {
 
 # This is partially adapted from the following SO thread
 # https://stackoverflow.com/questions/41668588/regex-to-match-scientific-notation
-_NUMB_PATTERN = r"^[+/-]?((?:0|[1-9]\d*)(?:\.\d*)?(?:[eE][+\-]?\d+)|\d*\.?\d+|\d+\.?\d*|nan\s|inf\s)"  # noqa: E501
+_NUMB_PATTERN = (
+    r"[+/-]?(?:((?:\d\.?\d*[Ee][+\-]?\d+)|(?:\d+\.\d*|\d*\.\d+))|\d+|inf\s|nan\s)"
+)
 # *all* greek letters are considered valid unit string elements.
 # This may be an overshoot. We rely on unyt.Unit to do the actual validation
 _UNIT_PATTERN = r"([α-ωΑ-Ωa-zA-Z]+(\*\*([+/-]?[0-9]+)|[*/])?)+"
@@ -1376,7 +1378,7 @@ class unyt_array(np.ndarray):
         --------
         >>> from unyt import unyt_quantity
         >>> unyt_quantity.from_string("1cm")
-        unyt_quantity(1., 'cm')
+        unyt_quantity(1, 'cm')
         >>> unyt_quantity.from_string("+1e3 Mearth")
         unyt_quantity(1000., 'Mearth')
         >>> unyt_quantity.from_string("-10. kg")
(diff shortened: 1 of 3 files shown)


ngoldbaum:
argh, sorry! didn't mean to push this upstream, fixing...

ngoldbaum:
Again, sorry for the noise, that's what I get for trying to do git stuff after getting a covid booster.

neutrinoceros:
Oops, I tried to request a review from myself before I saw you already did it (mobile interface is confusing). I will take a look shortly, but I can already say that the change of behavior you're proposing for `unyt_quantity.from_string` seems both desirable and acceptable to me.

unyt/array.py
```diff
@@ -1376,30 +1378,37 @@ def from_string(s, unit_registry=None):
...
-        res = re.search(_NUMB_REGEXP, v)
-        num = res.group()
-        res = re.search(_UNIT_REGEXP, v[res.span()[1] :])
-        unit = res.group()
-        return float(num) * Unit(unit, registry=unit_registry)
+        else:
+            res = re.search(_NUMB_REGEXP, v)
+            num = res.group()
+            res = re.search(_UNIT_REGEXP, v[res.span()[1] :])
+            unit = res.group()
```
neutrinoceros, line 1406: This is a nit, I'm trying to avoid reusing variable names for different types so we can maybe start doing static type-checking without too much rewrites in the future
```suggestion
            snum = re.search(_NUMB_REGEXP, v).group()
            sunit = re.search(_UNIT_REGEXP, v[res.span()[1] :]).group()
```

unyt/array.py
```diff
@@ -1376,30 +1378,37 @@ def from_string(s, unit_registry=None):
...
+        try:
+            num = int(num)
+        except ValueError:
+            num = float(num)
```
neutrinoceros, line 1410: (baring in mind that mixing `float` and `int` is something type checkers are very indulgent about)
```suggestion
            num = int(snum)
        except ValueError:
            num = float(snum)
```

unyt/tests/test_unyt_array.py
```diff
@@ -2560,42 +2560,43 @@ def test_string_formatting():
...
+        ("+inf g", float("inf") * Unit("g"), "inf g"),
+        ("-inf g", -float("inf") * Unit("g"), "-inf g"),
+        ("1", 1.0 * Unit(), "1 dimensionless"),
+        ("g", 1.0 * Unit("g"), "1 g"),
     ],
 )
-def test_valid_quantity_from_string(s, expected):
+def test_valid_quantity_from_string(s, expected, normalized):
     actual = unyt_quantity.from_string(s)
     if "nan" in s:
```
neutrinoceros, line 2595: looking back I think I should have separated these special cases into a different test. Not required here though.

neutrinoceros (approved):
LGTM. Just have a couple nits to reports, but nothing blocking. Thank you for doing this Nathan

PR #280, opened 2022-10-01
Improve test coverage

ngoldbaum:
This PR improves a few places where we were missing test coverage. There are only two areas where we don't have 100% test coverage after this, both of which are hard to test and I'm OK with not having test coverage there.

The fixes fall into two flavors:

* Removing unnecessary compatibility code for old versions of dependencies we no longer support
* Fixing places where new code that's been added in the past couple years didn't have test coverage, either by removing unnecessary code or fixing existing broken tests.

@chrishavlin one of the changes was to delete `unyt_dask_array.__array_function__` since none of the existing tests invoke that function. Can you confirm that my change is correct and `__array_function__` is never called by code outside of `unyt`? It could also be that the existing tests don't invoke the necessary preconditions for `__array_function__` to get called too.

I also noticed that the docstring for `unyt_array.__pow__` that @JBorrow added included a lot of implementation details that users probably don't care about, so I moved that content into a comment.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1701,10 +1701,7 @@ class unyt_array(np.ndarray):
         ret = super().__getitem__(item)
         if getattr(ret, "shape", None) == ():
             ret = unyt_quantity(ret, bypass_validation=True, name=self.name)
-        try:
             ret.units = self.units
-        except AttributeError:
-            pass
         return ret
 
     def __pow__(self, p, mod=None, /):
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -15,6 +15,7 @@ unyt_array class.
 
 import copy
 import re
+import warnings
 from functools import lru_cache
 from numbers import Number as numeric_type
 
@@ -107,8 +108,6 @@ from numpy import (
     trunc,
 )
 from numpy.core.umath import _ones_like, clip
-import warnings
-
 from sympy import Rational
 
 from unyt._on_demand_imports import _astropy, _dask, _pint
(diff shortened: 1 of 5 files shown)


neutrinoceros:
Quick question: where do we set or document our target minimal supported versions for optional dependencies like matplotlib ?
My best guess is that we can't claim support for versions older than 2.2.5 because that's the oldest one that shipped wheels for Python 3.8, our current minimal requirement.

ngoldbaum:
It's not explicitly mentioned in the docs. I go by the versions specified in the `py38-versions` tox job:

https://github.com/yt-project/unyt/blob/43612b60d08e499f5715ebec5432d9ee25311f4a/tox.ini#L29-L39

chrishavlin:
I'll check on the '__array_function__' question. It's possible that it shouldn't be deleted but if that's the case I'll also write up a test for it.

ngoldbaum:
@chrishavlin here's a test script that does trigger `__array_function__`, you need to write a test that uses a numpy function to do something with a `unyt_dask_array`:

```
from unyt import dask_array as uda
import dask.array as da
import numpy as np

x = da.arange(10000, chunks=(1000,))
x_da = uda.unyt_from_dask(x, 'm')
np.any(x_da)
```

Would you mind writing some tests to see if various numpy functions do the right thing? I'm happy to pull from you, we can also followup in a future PR.

chrishavlin:
Wrote some new tests to catch the unyt-dask array `__array_function__`. I pushed them to a new branch on my fork (`dask_array_func_tests`, https://github.com/chrishavlin/unyt/tree/dask_array_func_tests ). Happy to submit as a follow up PR, or you can pull in those changes here. Whatever is easiest, just let me know!

PR #284, opened 2022-10-09
replace versioneer with setuptools_scm

### LINKED ISSUE #254: Migrate away from versioneer ?

ngoldbaum:
Closes #254.

I did some profling and saw slowdowns about half as large as what @neutrinoceros reported. My first try at this was to see if updating versioneer helped. It turns out the latest version is even slower.

setuptools_scm lets us jettison a bunch of vendored code that we've never bothered to update. So far it hasn't broken but it probably will in the future as the python language changes. `setuptools_scm` seems to be a drop-in replacement too, I don't notice any real differences from a developer perspective, I still get sensible versions based on the last tag and git version in the `__version__` attribute, it just gets generated at build time.


pyproject.toml
```diff
@@ -26,3 +24,9 @@ exclude = '''
 profile = "black"
 combine_as_imports = true
 skip =  ["venv", "benchmarks"]
+
+[build-system]
+requires = ["setuptools>=45", "setuptools_scm[toml]>=6.2", "wheel"]
```
neutrinoceros, line 29: This is a fundamental part of a `pyproject.toml` file, and is traditionally the very first section. Could you move it to the top ?

pyproject.toml
```diff
@@ -26,3 +24,9 @@ exclude = '''
...
+
+[tool.setuptools_scm]
+write_to = "unyt/_version.py"
```
neutrinoceros, line 32: I suggest using the `"post-release"` version scheme instead of the default
```suggestion
write_to = "unyt/_version.py"
version_scheme = "post-release"
```

This gives for instance `'2.9.2.post105+g7883d04.d20221010'` instead of `'2.9.3.dev105+g7883d04'`, which I think is preferable because we don't know if the next release is going to be numbered `2.9.3` yet

neutrinoceros (commented):
I compared this with my own previous attempt (#255).
It seems we could also clean up `.pre-commit-config.yaml`

neutrinoceros (approved):

PR #288, opened 2022-10-10
MNT: fixup pre-commit-config.yaml

neutrinoceros:
#284 broke pre-commit such that all files are now being excluded. This fixes that


neutrinoceros:
pre-commit.ci autofix

Issue #289, opened 2022-10-10
NEP 18 progress tracker


neutrinoceros:
We want to implement NEP 18 progressively on main. #200 will be the first step towards that goal. This issue will be used to keep track of what's left to do in this project.

The full list of functions that need re-implementation is found here https://github.com/numpy/numpy/issues/15544#issuecomment-584453408

Here it is, converted to a TODO list (the `numpy.` namespace is implicit everywhere)

<details> <summary> TODO </summary>

- [ ] allclose
- [ ] amax
- [ ] amin
- [ ] angle
- [ ] any
- [ ] append
- [ ] apply_along_axis
- [ ] apply_over_axes
- [ ] argmax
- [ ] argmin
- [ ] argpartition
- [ ] argsort
- [ ] argwhere
- [ ] around
- [ ] array_equal
- [ ] array_equiv
- [ ] array_repr
- [ ] array_split
- [ ] array_str
- [ ] asfarray
- [ ] asscalar
- [ ] atleast_1d
- [ ] atleast_2d
- [ ] atleast_3d
- [ ] average
- [ ] bincount
- [ ] block
- [ ] broadcast_arrays
- [ ] broadcast_to
- [ ] busday_count
- [ ] busday_offset
- [ ] can_cast
- [ ] choose
- [ ] clip
- [ ] column_stack
- [ ] common_type
- [ ] compress
- [ ] convolve
- [ ] copy
- [ ] copyto
- [ ] corrcoef
- [ ] correlate
- [ ] count_nonzero
- [ ] cov
- [ ] cumprod
- [ ] cumproduct
- [ ] cumsum
- [ ] datetime_as_string
- [ ] delete
- [ ] diag
- [ ] diag_indices_from
- [ ] diagflat
- [ ] diagonal
- [ ] diff
- [ ] digitize
- [ ] dsplit
- [ ] dstack
- [ ] ediff1d
- [ ] einsum
- [ ] einsum_path
- [ ] empty_like
- [ ] expand_dims
- [ ] extract
- [ ] fft.fft
- [ ] fft.fft2
- [ ] fft.fftn
- [ ] fft.fftshift
- [ ] fft.hfft
- [ ] fft.ifft
- [ ] fft.ifft2
- [ ] fft.ifftn
- [ ] fft.ifftshift
- [ ] fft.ihfft
- [ ] fft.irfft
- [ ] fft.irfft2
- [ ] fft.irfftn
- [ ] fft.rfft
[...]

PR #293, opened 2022-10-11
ENH: implement 8 new NEP 18 functions

neutrinoceros:
implements https://github.com/yt-project/unyt/issues/289#issuecomment-1274964636

TODO:
- [x] get CI green for existing tests
- [x] add new tests

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -24,7 +24,7 @@ deps =
     dask[array,diagnostics]
     setuptools_scm
 commands =
-    pytest --cov=unyt --cov-append --doctest-modules --doctest-plus --doctest-rst --basetemp={envtmpdir} -W once
+    pytest --cov=unyt --cov-append --doctest-modules --doctest-plus --doctest-rst --basetemp={envtmpdir}
     coverage report --omit='.tox/*'
 
 [testenv:py38-versions]
@@ -61,7 +61,7 @@ depends = begin
 commands =
     # don't do doctests in rst files due to lack of way to specify optional
     # test dependencies there
-    pytest --cov=unyt --cov-append --doctest-modules --doctest-plus --basetemp={envtmpdir} -W once
+    pytest --cov=unyt --cov-append --doctest-modules --doctest-plus --basetemp={envtmpdir}
     coverage report --omit='.tox/*'
 
 [testenv:py38-docs]
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -189,10 +189,10 @@ def histogramdd(
 
 
 def _validate_units_consistency(arrs):
-    from unyt.array import unyt_array
-
-    if not all(isinstance(a, unyt_array) for a in arrs):
-        raise RuntimeError("Not all of your arrays are unyt_arrays.")
+    # NOTE: we cannot validate that all arrays are unyt_arrays
+    # by using this as a guard clause in unyt_array.__array_function__
+    # because it's already a necessary condition for numpy to use our
+    # custom implementations
     a1 = arrs[0]
     if not all(a.units == a1.units for a in arrs[1:]):
         raise RuntimeError("Your arrays must have identical units.")
@@ -211,7 +211,7 @@ def concatenate(arrs, /, axis=0, out=None, dtype=None, casting="same_kind"):
 def cross(a, b, axisa=-1, axisb=-1, axisc=-1, axis=None):
     prod_units = a.units * b.units
     return (
-        np.dot._implementation(
+        np.cross._implementation(
             a.ndview, b.ndview, axisa=axisa, axisb=axisb, axisc=axisc, axis=axis
         )
         * prod_units
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -1,7 +1,7 @@
 import numpy as np
 from packaging.version import Version
 
-from unyt.exceptions import UnitConversionError, UnitOperationError
+from unyt.exceptions import UnitConversionError, UnitInconsistencyError
 
 NUMPY_VERSION = Version(np.__version__)
 _HANDLED_FUNCTIONS = {}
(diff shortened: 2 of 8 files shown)


tox.ini
```diff
@@ -24,15 +24,15 @@ deps =
     dask[array,diagnostics]
     setuptools_scm
 commands =
-    pytest --cov=unyt --cov-append --doctest-modules --doctest-plus --doctest-rst --basetemp={envtmpdir} -W once
+    pytest --cov=unyt --cov-append --doctest-modules --doctest-plus --doctest-rst --basetemp={envtmpdir}
```
ngoldbaum, line 27: Can you explain a little more why you made this change?

tox.ini
```diff
@@ -24,15 +24,15 @@ deps =
...
     coverage report --omit='.tox/*'
 
 [testenv:py38-versions]
 deps =
     docutils
     pytest
     sympy==1.5
-    numpy==1.17.5
+    numpy==1.20.0
```
ngoldbaum, line 35: I think we can’t bump to such a recent numpy, even if we drop support for python 3.9 we’d still need to support numpy 1.19.

Why did you bump the minimum version? The commit logs talk about `np.concatenate` - couldn’t you put that in an import guard or something similar?

unyt/_array_functions.py
```diff
@@ -186,3 +186,83 @@ def histogramdd(
...
+
+
+def _validate_units_consistency(arrs):
+    # NOTE: we cannot validate that all arrays are unyt_arrays
+    # by using this as a guard clause in unyt_array.__array_function__
+    # because it's already a necessary condition for numpy to use our
+    # custom implementations
+    a1 = arrs[0]
+    if not all(a.units == a1.units for a in arrs[1:]):
+        raise RuntimeError("Your arrays must have identical units.")
```
ngoldbaum, line 198: Should this be `UnitOperationError` or one of the other unyt-specific errors?

unyt/array.py
```diff
@@ -2214,15 +2214,16 @@ def uconcatenate(arrs, axis=0):
     >>> from unyt import cm
     >>> A = [1, 2, 3]*cm
     >>> B = [2, 3, 4]*cm
-    >>> uconcatenate((A, B))
+    >>> uconcatenate((A, B)) # doctest: +IGNORE_EXCEPTION_DETAIL
+    Traceback (most recent call last):
+    DeprecationWarning: ...
```
ngoldbaum, line 2219: Including the deprecation warning details in the doctest is gross but might be unavoidable. I can try poking at this too. I tell myself that running the doctests is worth the occasional uglyness that shows up.

ngoldbaum (commented):
Thanks for plugging at this. Mostly looks good but I have a couple places I need clarification.

tox.ini
neutrinoceros: I found it was necessary in order to get all deprecation warnings poping up in doctests instead of just the first one calling `warn_deprecated`

tox.ini
neutrinoceros: `np.concatenate`'s signature changed in numpy 1.20 (the last 2 arguments weren't available before that), so my first implementation for it wasn't compatible with numpy<1.20

I bumped because I was looking for the quicker fix to get CI back to green, but I can refine this now and provide two different implementations, with a runtime check of numpy's version using `packaging` (this would be a new hard dependency, but a cheap one: it's almost certainly an indirect dependency to unyt already)

unyt/_array_functions.py
neutrinoceros: You tell me. Here I basically copied and adapted the content of `unyt.array._validate_numpy_wrapper_units`. I'm happy to change this to any custom error you deem more appropriate.

unyt/array.py
neutrinoceros: I couldn't find a more elegant solution yet, but I figure "keeping docstrings pretty in deprecated functions" might not be worth the effort (assuming it's even possible).
Happy to simplify this if you have an idea how to, however.

tox.ini
ngoldbaum: Adding `packaging` is fine.

unyt/_array_functions.py
ngoldbaum: I think both should probably be `UnitOperationError`, if the other one is a `RuntimeError` that's an oversight. Ideally users shouldn't need to guess which errors `unyt` will generate for a given operation, and `UnitOperationError` is what we raise elsewhere when a user tries to do an operation with invalid units.

tox.ini
neutrinoceros: thanks ! updated

unyt/tests/test_array_functions.py
```diff
@@ -199,3 +199,83 @@ def test_histogramdd():
...
+    x2 = np.random.normal(size=100) * cm
+    res = np.concatenate((x1, x2))
+    assert 1 * res.units == 1 * cm
+    assert res.shape == (200,)
+
+
+def test_concatenate_different_units():
+    x1 = np.random.normal(size=100) * cm
+    x2 = np.random.normal(size=100) * s
+    with pytest.raises(RuntimeError, match=r"Your arrays must have identical units\."):
```
ngoldbaum, line 215: Forgot to update the error here. FWIW I strongly encourage you to run the tests locally before you push to github, it's pretty fast.

ngoldbaum: https://unyt.readthedocs.io/en/stable/contributing.html#testing-unyt

unyt/_array_functions.py
neutrinoceros: Ah, I was a little careless when I did this and didn't realise that `UnitOperationError.__init__` had a template error message. Because it refers to an "operator", I'm not sure the wording works that well in the context of array functions. Also, `UnitOperationError` takes 2 units at most, but here we might have an arbitrary number of them. Should I subclass it to a more finely tailored `UnitInconstencyError` class ?

unyt/tests/test_array_functions.py
neutrinoceros: I thought I did actually... anyway it's even more broken than just this (see above thread)

unyt/_array_functions.py
ngoldbaum: Sure, that make sense. Sorry, didn't realize that wouldn't just be a straightforward substitution.

neutrinoceros: no worries

PR #294, opened 2022-10-12
Add common energy units

samgdotson:
This PR adds the following units common in energy systems engineering
* MMBTU (1 million BTU)
* therm (100,000 BTU)
* quad (1e15 BTU)
* Wh (prefixable)

Additionally, this PR redefines a BTU using a J_per_BTU physical ratio -- eliminating a "magic number."

There is some ambiguity around "MBTU" being 1000 or 1,000,000 BTU. Happy to include this prefix for completeness or to mention this ambiguity in the docs (and that users are welcome to define their own MBTU unit).


unyt/_physical_ratios.py
```diff
@@ -122,6 +122,11 @@
 kelvin_per_rankine = 5.0 / 9.0
 watt_per_horsepower = 745.69987158227022
 erg_per_s_per_watt = 1e7
+J_per_BTU = 1055.0559
+watt_per_hour = 3600
```
ngoldbaum, line 126: this should be called joules_per_watt_hour I think? W/hr isn't dimensionless so calling it that doesn't make sense in this file, which only contains dimensionless ratios.

unyt/_unit_lookup_table.py
```diff
@@ -544,7 +553,29 @@
         ("hp", ("horsepower",)),
         ("oz", ("ounce",)),
         ("cal", ("calorie",)),
-        ("BTU", ("british_thermal_unit",)),
+        (
+            "BTU",
+            (
+                "british_thermal_unit",
+                "btu",
```
ngoldbaum, line 560: Good catch on this alias, I didn't realize the automatic aliases we create to handle capitalization in unit names explicitly excludes all-upper-case names. It makes sense it works that way though, as well as adding the explicit aliases you're adding here.

unyt/_unit_lookup_table.py
```diff
@@ -544,7 +553,29 @@
...
+            ),
+        ),
+        (
+            "MMBTU",
+            (
+                "million_british_thermal_unit",
```
ngoldbaum, line 566: I don't think we should support this alias as I doubt anyone would ever try to type it. Happy to be persuaded otherwise if you disagree.

unyt/_unit_lookup_table.py
```diff
@@ -195,7 +200,11 @@
             ),
         ),
         ("cal", (4.184, dimensions.energy, 0.0, r"\rm{cal}", True)),
-        ("BTU", (1055.0559, dimensions.energy, 0.0, r"\rm{BTU}", False)),
+        ("BTU", (J_per_BTU, dimensions.energy, 0.0, r"\rm{BTU}", False)),
+        ("MMBTU", (btu_per_mmbtu, dimensions.energy, 0.0, r"\rm{MMBTU}", False)),
+        ("therm", (btu_per_therm, dimensions.energy, 0.0, r"\rm{therm}", False)),
+        ("quad", (btu_per_quad, dimensions.energy, 0.0, r"\rm{quad}", False)),
```
ngoldbaum, line 206: If anyone else is unfamiliar with these units, from wikipedia:

>Units kBtu are used in building energy use tracking and heating system sizing. Energy Use Index (EUI) represents kBtu per square foot of conditioned floor area. "k" stands for 1,000.

>The unit Mbtu is used in natural gas and other industries to indicate 1,000 BTUs.[[11]](https://en.wikipedia.org/wiki/British_thermal_unit#cite_note-EIA-13)[[12]](https://en.wikipedia.org/wiki/British_thermal_unit#cite_note-Price-14) However, there is an ambiguity in that the metric system (SI) uses the prefix "M" to indicate '[Mega-](https://en.wikipedia.org/wiki/Mega-)', one million (1,000,000). Even so, "MMbtu" is often used to indicate one million BTUs particularly in the oil and gas industry.[[13]](https://en.wikipedia.org/wiki/British_thermal_unit#cite_note-APS-15)

>Energy analysts accustomed to the metric "k" ('[kilo-](https://en.wikipedia.org/wiki/Kilo-)') for 1,000 are more likely to use MBtu to represent one million, especially in documents where M represents one million in other energy or cost units, such as MW, MWh and $.

>The unit '[therm](https://en.wikipedia.org/wiki/Therm)' is used to represent 100,000 BTUs.[[11]](https://en.wikipedia.org/wiki/British_thermal_unit#cite_note-EIA-13) A decatherm is 10 therms or one MMBtu (million Btu). The unit [quad](https://en.wikipedia.org/wiki/Quad_(unit)) is commonly used to represent one quadrillion (1015) BTUs.[[13]](https://en.wikipedia.org/wiki/British_thermal_unit#cite_note-APS-15)

Definitely awesome that there's some ambiguity about whether SI prefixes make sense or not. I agree with your choice to just not allow it.

unyt/_unit_lookup_table.py
```diff
@@ -195,7 +200,11 @@
...
+        ("Wh", (watt_per_hour, dimensions.energy, 0.0, r"\rm{Wh}", True)),
```
ngoldbaum, line 207: According to wikipedia, allowing SI prefixes for this one is also a bit controversial. In the interest of practicality I think it's fine.

unyt/_unit_lookup_table.py
samgdotson: I'm not attached to it. I only added it for symmetry with BTU.

unyt/_physical_ratios.py
samgdotson: Yep, that makes a lot more sense!

PR #297, opened 2022-10-15
add test to make sure we're testing all wrappable numpy functions

ngoldbaum:
Obviously we haven't added support for all wrappable functions yet, so I added an explicit whitelist. If this is merged you'll need to remove functions you add wrappers for from `NOT_HANDLED_FUNCTIONS`. Doing this ensures we eventually wrap everything we want to wrap. It also ensures we catch it when numpy adds new functions that can be wrapped in the future.


ngoldbaum:
I also realized after fixing the tests this also prevents us from adding unnecessary wrappers for functions that have already been deprecated and removed from numpy

ngoldbaum:
@neutrinoceros you added a wrapper for matmul, but matmul doesn't yet support wrapping, so I removed that implementation. It does happen to work as one might expect with `__unyt_array__`, so I left the test in that checks that.

neutrinoceros:
Sounds great. I will review this closely tomorrow. Thank you !

ngoldbaum:
Ah, of course, the reason matmul isn't getting picked up is because it's a ufunc and it uses `_array_ufunc`.

neutrinoceros:
> Ah, of course, the reason matmul isn't getting picked up is because it's a ufunc and it uses _array_ufunc.

🤦🏻‍♂️ that makes sense. Welp, I agree with everything you did here. I'll approve now and can merge later today if you don't need to iterate more.

neutrinoceros (approved):

PR #298, opened 2022-10-16
support ndarray as arguments for wrapped functions

ngoldbaum:
See https://github.com/yt-project/unyt/issues/289#issuecomment-1279862821 for motivation.  With this PR the script in the comment after that one prints:

```
[[ 6. 12. 18.]
 [ 6. 12. 18.]
 [ 6. 12. 18.]] g*kg
[[ 6. 12. 18.]
 [ 6. 12. 18.]
 [ 6. 12. 18.]]
<class 'unyt.array.unyt_array'> <class 'numpy.ndarray'>
[[ 6 12 18]
 [ 6 12 18]
 [ 6 12 18]] g*kg
[[ 6 12 18]
 [ 6 12 18]
 [ 6 12 18]]
<class 'unyt.array.unyt_array'> <class 'numpy.ndarray'>
```

(I guess not great that there's still a difference in the result's dtype..)

Happy to add more tests if you want me to.

### DIFF
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -345,6 +345,37 @@ def test_dot_matrices(out):
         assert out.units == res.units
 
 
+def test_dot_mixed_ndarray_unyt_array():
+    a = np.ones((3, 3))
+    b = np.ones((3, 3)) * cm
+
+    res = np.dot(a, b)
+
+    assert isinstance(res, unyt_array)
+    assert res.units == cm
+
+    out = np.zeros((3, 3))
+    res = np.dot(a, b, out=out)
+
+    assert isinstance(res, unyt_array)
+    assert type(out) is np.ndarray
+    np.testing.assert_array_equal(out, res)
+
+    out = np.zeros((3, 3)) * km
+    res = np.dot(a, b, out=out)
+
+    assert isinstance(res, unyt_array)
+    assert isinstance(out, unyt_array)
+    assert res.units == out.units == km
+
+    out = np.zeros((3, 3)) * km
+    res = np.dot(b, a, out=out)
+
+    assert isinstance(res, unyt_array)
+    assert isinstance(out, unyt_array)
+    assert res.units == out.units == km
+
+
 def test_invalid_dot_matrices():
     a = np.arange(9) * cm
     a.shape = (3, 3)
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -335,7 +335,9 @@ def test_dot_matrices(out):
 
     if out is not None:
         np.testing.assert_array_equal(res, out)
-        assert res is out
+        assert np.shares_memory(res, out)
+        assert isinstance(res, unyt_array)
+        assert isinstance(out, np.ndarray)
 
     if isinstance(out, unyt_array):
         # check that the result can be converted to predictible units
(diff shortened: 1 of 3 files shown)


unyt/tests/test_array_functions.py
```diff
@@ -50,14 +50,47 @@ def test_dot_matrices(out):
...
+
+    assert isinstance(res, unyt_array)
+    assert res.units == cm
+
+    out = np.zeros((3, 3))
+    res = np.dot(a, b, out=out)
+
+    assert isinstance(res, unyt_array)
+    assert type(out) is np.ndarray
+    np.testing.assert_array_equal(out, res)
```
neutrinoceros, line 77: I don't know *if* we can pass this too, but it'd be great to check that memory usage isn't doubled
(this is debatable but at the moment I believe it'd be the preferable behaviour, provided it's possible to implement. In any case this is the kind of subtle gotcha we'll need to document)
```suggestion
    np.testing.assert_array_equal(out, res)
    assert out.base is res
```

unyt/tests/test_array_functions.py
```diff
@@ -50,14 +50,47 @@ def test_dot_matrices(out):
...
+
+    out = np.zeros((3, 3)) * km
+    res = np.dot(a, b, out=out)
+
+    assert isinstance(res, unyt_array)
+    assert isinstance(out, unyt_array)
+    assert res.units == out.units == km
+
```
neutrinoceros, line 85:
```suggestion
    assert res.units == out.units == km
    assert res is out
```

unyt/tests/test_array_functions.py
```diff
@@ -50,14 +50,47 @@ def test_dot_matrices(out):
...
+    out = np.zeros((3, 3)) * km
+    res = np.dot(b, a, out=out)
```
neutrinoceros, line 87:
```suggestion
    # inverting a and b
    res = np.dot(b, a, out=out)
```

unyt/tests/test_array_functions.py
```diff
@@ -50,14 +50,47 @@ def test_dot_matrices(out):
...
+
+    assert isinstance(res, unyt_array)
+    assert isinstance(out, unyt_array)
+    assert res.units == out.units == km
+
```
neutrinoceros, line 92:
```suggestion
    assert res.units == out.units == km
    assert res is out
```

neutrinoceros (commented):
I agree this is the way to go, thank you very much for pointing it out and working out the solution.

My only concern with this is that we should specify the relationship between the out array and the return value if they are not the same object (i.e. when `out` is an `ndarray` and the result array is a `unyt_array`): should they own their data (double the memory usage) or share it ? I think the return array shouldn't own its data, because that's the closest thing we can do to "out and return value are the same object" here, and it seems safe to avoid doing copies in this context.
In pure numpy, I'm not even sure that it makes much sense to use `out` *and* capture the return value, but since we're making a distinction between the two, we're also creating an incentive for such usage, so I think we better make sure we know what we're doing.

ngoldbaum:
Rather than messing with `base`, I used `np.shares_memory` to check if the arrays share data. They do share data in all cases, there's no doubling of memory.

unyt/tests/test_array_functions.py
neutrinoceros: resolved with `numpynp.shares_memory`

neutrinoceros (approved):
Cool, thanks for using `np.shares_memory` (I didn't know about it), I agree it is cleaner than messing with `.base`

PR #304, opened 2022-10-17
make array functions with out arguments behave more similarly to ufuncs

ngoldbaum:
This proposes another change to how we handle `out` arguments, again to make the behavior more similar to the current behavior of ufuncs. To illustrate this, take a look at the following script:

```
In [1]: import unyt as u

In [2]: import numpy as np

In [3]: a = [1, 2, 3]*u.g

In [4]: b = [1, 2, 3]*u.kg

In [5]: c = np.zeros((3, 3))*u.W

In [6]: np.add(a, b, out=c)
Out[6]: 
unyt_array([[1001., 2002., 3003.],
            [1001., 2002., 3003.],
            [1001., 2002., 3003.]], 'g')

In [7]: print(c)
[[1001. 2002. 3003.]
 [1001. 2002. 3003.]
 [1001. 2002. 3003.]] g
```

From which we see that ufuncs overwrite the output units.

IMO this is preferable to the current behavior for e.g. np.dot, which will check if `out` has units, raise an error if the units aren't convertible, and convert the result to `out`'s units otherwise. We don't actually need to do all that conversion work. We also IMO don't need to raise an error unnecessarily if we can just do what the user is asking us to do and make `out` have the same data as the result of the operation.

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -2,7 +2,7 @@ import numpy as np
 from packaging.version import Version
 
 from unyt.array import NULL_UNIT, unyt_array
-from unyt.exceptions import UnitConversionError, UnitInconsistencyError
+from unyt.exceptions import UnitInconsistencyError
 
 NUMPY_VERSION = Version(np.__version__)
 _HANDLED_FUNCTIONS = {}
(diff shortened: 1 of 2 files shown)


neutrinoceros:
In principle I still believe an error _should_ be raised if the out array has incompatible units, *but* that said, api-consistency beats purity, and I don't wish to break more downstream code than necessary with the implementation of NEP 18 so I can get behind this.

unyt/_array_functions.py
```diff
@@ -26,39 +26,23 @@ def array2string(a, *args, **kwargs):
...
-    except UnitConversionError as exc:
-        raise TypeError(
-            "output array is not acceptable "
-            f"(units '{out.units}' cannot be converted to '{units}')"
-        ) from exc
-
-
-@implements(np.dot)
-def dot(a, b, out=None):
+def product_helper(a, b, out, func):
```
neutrinoceros, line 29: I think this should still be a hidden function
```suggestion
def _product_helper(a, b, out, func):
```

neutrinoceros (commented):
just got one minor request

unyt/_array_functions.py
ngoldbaum: It’s already in a private namespace

neutrinoceros: good point

PR #305, opened 2022-10-17
add liter to the default unit lookup table

ngoldbaum:
I can't believe this hasn't been added until now. Whoops.


neutrinoceros (approved):
🤯

PR #306, opened 2022-10-18
ENH: (NEP 18) allow pure ndarrays to match any units in _validate_units_consistency

neutrinoceros:
In order to get more inline with ufuncs (follow up to #302 and #304)
Also tests 2 basically free functions that already work.


neutrinoceros:
note that the `np.append` test I added only works with the change that I made to `_validate_units_consistency`

unyt/tests/test_array_functions.py
```diff
@@ -654,3 +654,15 @@ def test_trim_zeros():
...
+
+
+def test_any():
+    assert not np.any([0, 0, 0] * cm)
+    assert np.any([1, 0, 0] * cm)
+
+
+def test_append():
+    a = [0, 1, 2, 3] * cm
+    b = np.append(a, [4, 5, 6])
```
ngoldbaum, line 666: shouldn’t this error out unless `a` is dimensionless?

neutrinoceros: yes, but are we not treating (pure arrays / list / pick your poison) as dimensionless ?

ngoldbaum: Yup, which is why this should error out, since `a` has non-dimensionless units.

ngoldbaum: IMO, no. That would also be different from how ufuncs work.

neutrinoceros: fair enough.
I reverted the previous behaviour but still fixed the validator function because it doesn't behave properly on the main branch for pure arrays

unyt/_array_functions.py
```diff
@@ -172,9 +172,12 @@ def _validate_units_consistency(arrs):
     # by using this as a guard clause in unyt_array.__array_function__
     # because it's already a necessary condition for numpy to use our
     # custom implementations
-    ref_units = getattr(arrs[0], "units", NULL_UNIT)
-    if any(getattr(a, "units", NULL_UNIT) != ref_units for a in arrs[1:]):
-        raise UnitInconsistencyError(*(a.units for a in arrs))
```
neutrinoceros, line 177: the bug I'm addressing is that `a.units` escalates to an `AttributeError` for pure arrays

unyt/tests/test_array_functions.py
```diff
@@ -654,3 +654,27 @@ def test_trim_zeros():
...
+def test_any():
+    assert not np.any([0, 0, 0] * cm)
+    assert np.any([1, 0, 0] * cm)
+
+
+def test_append():
+    a = [0, 1, 2, 3] * cm
+    b = np.append(a, [4, 5, 6] * cm)
+    assert type(b) is unyt_array
+    assert 1 * b.units == 1 * cm
```
ngoldbaum, line 668: you can compare units directly, no need to multiply by one

neutrinoceros: Oh that's a relief, I don't know why this stuck in my mind. I'll do another PR to simplify my existing tests then

PR #309, opened 2022-10-19
ENH: (NEP 18) test 7 new functions, impl 2

neutrinoceros:


unyt/_array_functions.py
```diff
@@ -167,23 +167,35 @@ def histogramdd(
     return counts, tuple(_bin * u for _bin, u in zip(bins, units))
 
 
-def _validate_units_consistency(arrs):
+def get_units(arrays):
+    units = []
+    for sub in arrays:
+        if isinstance(sub, np.ndarray):
```
neutrinoceros, line 173: this check is not as general as it should, we need to capture all non-nested arraylike objects here, which is a bit tricky

neutrinoceros: actually I'll take that back. `np.ndarray` instances are the *only* objects for which we know that we can grab a single unit without iterating over each elements.
This implementation will probably be dramatically slow when *nested lists of quantities* are passed but hey, I guess users shouldn't be doing that at scale _and_ expect the very best of performances, so all in all this seems reasonable to me

PR #310, opened 2022-10-31
TST: add Python 3.11 to test matrix

neutrinoceros:
Note that Python 3.11 is not yet available via pyenv, so I'm not updating dev docs for now.

PR #318, opened 2022-11-11
BLD: migrate static metadata from setup.py to pyproject.toml

neutrinoceros:
Trying to get static metadata more centralized into a single file, and adopt newly defined standards.

Recently, setuptools became capable of using _only_ `pyproject.toml`, so `setup.py` could be completely removed, but I'm leaving it for the one small task that cannot be converted to static metadata (combining README.rst+HISTORY.rst)

Two tools do not support pyproject.toml yet (flake8 and bumpversion) so I suggest keeping `setup.cfg` around for them, but in the future this file is meant to disappear completely.


ngoldbaum:
What do you think about adding a link in the readme to the changelog instead of concatenating the two files? I don’t think the marginal benefit of having the full changelog on pypi is worth having to think about the setup.py file at all.

neutrinoceros:
Good point. I researched the existing standards for this and
added a `changelog` and a `documentation` link in `[project.urls]`, which is supported by PyPI and should produce this result:
Screenshot 2022-11-12 at 09 57 40

neutrinoceros:
Note that we could also get rid of tox.ini and migrate its content into pypeoject.toml too, but I'll leave this to a future PR to avoid creating conflicts with #310

ngoldbaum:
I’d personally rather leave tox.ini as-is since it’s already pretty complicated.

neutrinoceros:
That's fine by me, I agree that the size of it justifies it being a separate file

ngoldbaum:
When we do the next release we should try to upload on the test pypi instance first to see if everything still works as expected.

PR #327, opened 2022-11-22
 BLD: remove wheel from explicit build time requirements (not needed in Python>=3.8)

neutrinoceros:
context: explictly declaring `wheel` as a build-time dependency is not needed with `pip>=19.0`, and the oldest supported Python version is now 3.8.0, which ships with `pip==19.3.2`


ngoldbaum:
This is news to me, can you share a link where I can read more about this change?

Do I still need `wheel` installed to generate a wheel locally via e.g. `python -m build`?

PR #331, opened 2022-12-04
TST: pin tox<4.0 to workaround incompatibility with tox-pyenv

neutrinoceros:
This PR isn't meant to be merged. Rather, I intend to use it as a playground to discover bugs with the upcoming new major release of tox. Best case scenario nothing breaks and we can close this.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,7 +39,7 @@ jobs:
       - name: Install Tox and any other packages
         run: |
           python -m pip install --upgrade pip setuptools wheel setuptools_scm
-          python -m pip install tox tox-pyenv coverage
+          python -m pip install "tox<4.0" tox-pyenv coverage # see https://github.com/tox-dev/tox-pyenv/issues/21
       - name: Build Docs
         if: matrix.python-version == '3.8'
         run: tox -e py38-docs -vvv


neutrinoceros:
turns out we *do* have an issue here, in that `tox-pyenv` depends on the `py` library, but doesn't declare it explicitly as a dependency. It *used* to work because `tox` itself had `py` as a dependency but it looks like it's going to be dropped in `tox` 4.0

I will report this upstream to `tox-pyenv`, but the project looks very much unmaintained (no updates at all in 5 years), so I'm don't have much hope of getting an answer there.

ngoldbaum:
I guess we could vendor it if it's unmaintained, I think the uses of `py` are pretty easy to patch out so we don't need to depend on that.

neutrinoceros:
Actually I have hopes it might get answered as one of the maintainers also works very actively on tox itself. Hopefully we don't need to vendor this anything 🤞

neutrinoceros:
tox 4.0 final was just released, I assume its too late to get the smooth transition I was hoping for
https://github.com/tox-dev/tox/releases/tag/4.0.0

neutrinoceros:
yup. Now CI is plain broken... I opened a cross ref on the main repo https://github.com/tox-dev/tox/issues/2611

neutrinoceros:
Going for it

PR #333, opened 2022-12-06
BUG: fix compatibility with numpy 1.25 (future)

neutrinoceros:
upstream test case from yt, see https://github.com/yt-project/yt/issues/4241
For further context, see https://github.com/numpy/numpy/issues/22744

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -137,6 +137,7 @@ from ._deprecation import warn_deprecated
 
 NULL_UNIT = Unit()
 POWER_MAPPING = {multiply: lambda x: x, divide: lambda x: 2 - x}
+DISALLOWED_DTYPES = ("S", "U", "a", "O", "M", "m", "b")
 
 __doctest_requires__ = {
     ("unyt_array.from_pint", "unyt_array.to_pint"): ["pint"],
@@ -251,21 +252,28 @@ def _bitop_units(unit1, unit2):
 
 def _coerce_iterable_units(input_object, registry=None):
     if isinstance(input_object, np.ndarray):
-        return input_object
-    if _iterable(input_object):
-        if any([isinstance(o, unyt_array) for o in input_object]):
+        ret = input_object
+    elif _iterable(input_object):
+        if any(isinstance(o, unyt_array) for o in input_object):
             ff = getattr(input_object[0], "units", NULL_UNIT)
-            if any([ff != getattr(_, "units", NULL_UNIT) for _ in input_object]):
+            if any(ff != getattr(_, "units", NULL_UNIT) for _ in input_object):
                 ret = []
                 for datum in input_object:
                     try:
                         ret.append(datum.in_units(ff.units))
                     except UnitConversionError:
                         raise IterableUnitCoercionError(str(input_object))
-                return unyt_array(np.array(ret), ff, registry=registry)
+                ret = unyt_array(np.array(ret), ff, registry=registry)
             # This will create a copy of the data in the iterable.
-            return unyt_array(np.array(input_object), ff, registry=registry)
-    return np.asarray(input_object)
+            else:
+                ret = unyt_array(np.array(input_object), ff, registry=registry)
+        else:
+            ret = np.asarray(input_object)
+    else:
+        ret = np.asarray(input_object)
+    if ret.dtype.char in DISALLOWED_DTYPES:
+        raise IterableUnitCoercionError(str(input_object))
+    return ret
 
 
 def _sanitize_units_convert(possible_units, registry):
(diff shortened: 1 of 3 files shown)


ngoldbaum:
I'm looking at fixing the bug right now actually. My plan was to push to this PR and merge it with the fix.

neutrinoceros:
fine by me ! it's actually getting pretty late in my time zone so, by all means !

neutrinoceros:
Also, given that it's known to break downstream code (yt), and we have no release date for unyt 3.0, I think it might be worth doing a bugfix release with just this fix, if it's small enough

neutrinoceros:
I'm game to do this !
What I had in mind was to branch out from the 2.9.2 tag. I'll manually go over the merge history to see if there's anything else worth backporting. My PyPI username is crobert

neutrinoceros:
Final note from me today but this is not going the way I had in mind; my understanding was that comparing a unyt quantity to a string _should_ be possible (the result would always be 0), and that's the expectation from yt's standpoint. If we're going to embrace it as invalid then it means yt needs a fix too.

unyt/array.py
```diff
@@ -265,7 +265,12 @@ def _coerce_iterable_units(input_object, registry=None):
                 return unyt_array(np.array(ret), ff, registry=registry)
             # This will create a copy of the data in the iterable.
             return unyt_array(np.array(input_object), ff, registry=registry)
-    return np.asarray(input_object)
+        ret = np.asarray(input_object)
+    else:
+        ret = np.asarray(input_object)
+    if ret.dtype.char in ("S", "U", "a", "O", "M", "m", "b"):
```
ngoldbaum, line 271: this tuple should probably be a module-level constant for slightly better performance

neutrinoceros:
Awesome, thank you very much !
About doing the release, is it safe for me to use test.pypi or do I need an other invite there ? (I'm not sure how frequently this index is purged, but it looks like unyt is currently not there)
just in case, my username _there_ is neutrinoceros

unyt/tests/test_unyt_array.py
```diff
@@ -1490,8 +1490,28 @@ def op_comparison(op, inst1, inst2, compare_class):
...
+
+def test_string_operations_raise_errors():
+    a = unyt_array([1, 2, 3], "g")
+    with pytest.raises(IterableUnitCoercionError):
         a + "hello"
+    with pytest.raises(IterableUnitCoercionError):
+        a * "hello"
+    with pytest.raises(IterableUnitCoercionError):
+        a ** "hello"
+    if Version(np.__version__).release[1] < 24:
```
neutrinoceros, line 1502: Just in case there's a numpy 2.x in the future
```suggestion
    if Version(np.__version__) < Version("1.24"):
```

unyt/exceptions.py
```diff
@@ -193,18 +193,20 @@ class IterableUnitCoercionError(Exception):
...
+    def __init__(self, op):
+        self.op = op
 
     def __str__(self):
         err = (
-            "Received a list or tuple of quantities with nonuniform units: "
-            "%s" % self.quantity_list
+            "Received an input or operand that cannot be converted "
+            "to a unyt_array with uniform units: "
+            "%s" % self.op
```
neutrinoceros, line 209: might as well
```suggestion
            f"to a unyt_array with uniform units: {self.op}"
```

unyt/array.py
```diff
@@ -265,7 +266,12 @@ def _coerce_iterable_units(input_object, registry=None):
                 return unyt_array(np.array(ret), ff, registry=registry)
```
neutrinoceros, line 266: While we're at it, might as well reduce the number of return statement to the bare minimum ?
```suggestion
                ret = unyt_array(np.array(ret), ff, registry=registry)
```

unyt/array.py
```diff
@@ -265,7 +266,12 @@ def _coerce_iterable_units(input_object, registry=None):
...
             # This will create a copy of the data in the iterable.
             return unyt_array(np.array(input_object), ff, registry=registry)
```
neutrinoceros, line 268: (to go with my previous suggestion)
```suggestion
            else:
                ret = unyt_array(np.array(input_object), ff, registry=registry)
```

unyt/tests/test_unyt_array.py
```diff
@@ -1490,8 +1490,28 @@ def op_comparison(op, inst1, inst2, compare_class):
...
+        with pytest.warns(FutureWarning):
+            assert a != "hello"
+    else:
+        assert (a != "hello").all()
+    q = 2 * g
```
neutrinoceros, line 1507: I'd have a slight preference for having 2 separated tests instead of just one, so we get more information in logs when something fails, ans this line seems to be a good opportunity to do that
```suggestion


def test_string_operations_raise_errors_quantity():
    q = 2 * g
```

neutrinoceros (commented):
I have a couple suggestions, but overall this is great.

PR #341, opened 2022-12-19
TST: fix a test to work on numpy 1.24

neutrinoceros:
Follow up to #333, fix a test that fails with the newly released version of numpy (1.24)


ngoldbaum:
I’m surprised the way the test is written now passes on old numpy versions that don’t raise a warning.

neutrinoceros:
But they do ! The test was already checking that actually, I changed the behaviour of the test for numpy 1.24+ only

PR #342, opened 2022-12-24
ENH: avoid importing dask when possible

neutrinoceros:
Calling `isinstance(i1, _dask.array.core.Array)` leads to importing dask if it's installed but not yet imported, but if it's not imported well, we don't need to check if i1 is a dask Array.
Changing this check to be more clever avoids forcing the import of dask (about as expensive as numpy) on `import yt.utilities.physical_constants` and avoids bloating yt's startup time.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -15,6 +15,7 @@ unyt_array class.
 
 import copy
 import re
+import sys
 import warnings
 from functools import lru_cache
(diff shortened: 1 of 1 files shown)


ngoldbaum:
@chrishavlin?

chrishavlin (approved):

PR #344, opened 2022-12-26
TST: migrate from tox-pyenv to tox-gh-actions

neutrinoceros:
Since tox-pyenv looks unmaintained (no response from the maintainer in 3 weeks now), let's experiment with a candidate replacement.
Because this makes tox 4 usable in CI for the first time, this may require some tweaks.
I'd also need to make changes in the dev guide if this works.

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -39,19 +39,6 @@ jobs:
       - name: Install Tox and any other packages
         run: |
           python -m pip install --upgrade pip setuptools wheel setuptools_scm
-          python -m pip install "tox<4.0" tox-pyenv coverage # see https://github.com/tox-dev/tox-pyenv/issues/21
-      - name: Build Docs
-        if: matrix.python-version == '3.8'
-        run: tox -e py38-docs -vvv
-      - name: Run Minimal Dependency Build
-        if: matrix.python-version == '3.8'
-        run: tox -e py38-dependencies
-      - name: Run Minimal Dependency Version Build
-        if: matrix.python-version == '3.8'
-        run: tox -e py38-versions
-      - name: Run Main Build
-        # Run tox using the version of Python in `PATH`
-        run: tox -e py -vvv
-      - name: Run In-place Tests
-        if: matrix.python-version == '3.8'
-        run: tox -e py38-unyt-module-test-function
+          python -m pip install tox tox-gh-actions
+      - name: Test
+        run: tox -vvv
--- a/CONTRIBUTING.rst
+++ b/CONTRIBUTING.rst
@@ -98,7 +98,7 @@ managing your python evironment using your operating system's package manager or
     $ pyenv install -s 3.8.13
     $ pyenv install -s 3.9.12
     $ pyenv install -s 3.10.4
-    $ pip install tox tox-pyenv
+    $ pip install tox
 
 4. Install your local copy into a virtualenv or conda environment. You can also
    use one of the python interpreters we installed using ``pyenv``::
--- a/tox.ini
+++ b/tox.ini
@@ -2,6 +2,12 @@
 envlist = py38-docs,begin,py38-dependencies,py38-versions,py{38,39,310},py38-unyt-module-test-function,end
 isolated_build = True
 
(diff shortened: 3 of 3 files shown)


neutrinoceros:
Only the `py38-docs` job is failing currently and it's not clear to me why. The only offender I see is a -probably harmless- warning raised at build time from setuptools_scm:
```
/opt/hostedtoolcache/Python/3.8.15/x64/lib/python3.8/site-packages/setuptools_scm/git.py:135: UserWarning: "/home/runner/work/unyt/unyt" is shallow and may cause errors
```

I'm not sure how to tell tox that it's okay.
For the sake of it, let's see if things run more smoothly with tox 3

neutrinoceros:
So this seems to only be a problem with tox 4; though it's a little hard to tell what gives since no version of `tox-gh-actions` currently supports both tox 3 and tox 4 so we can't just change one thing at a time.

neutrinoceros:
Actually the problem is much clearer when I reproduce locally:
```
$ python -m tox -e py38-docs

py38-docs: remove tox env folder /Users/robcleme/dev/yt-project/unyt/.tox/py38-docs
.pkg: remove tox env folder /Users/robcleme/dev/yt-project/unyt/.tox/.pkg
py38-docs: install_deps> python -I -m pip install 'dask[array,diagnostics]' 'matplotlib!=3.5.0' numpy pytest setuptools_scm sphinx sympy
.pkg: install_requires> python -I -m pip install 'setuptools>=61.2' 'setuptools_scm[toml]>=6.2'
.pkg: _optional_hooks> python /Users/robcleme/.pyenv/versions/unyt-dev-38/lib/python3.8/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: get_requires_for_build_sdist> python /Users/robcleme/.pyenv/versions/unyt-dev-38/lib/python3.8/site-packages/pyproject_api/_backend.py True setuptools.build_meta
.pkg: build_sdist> python /Users/robcleme/.pyenv/versions/unyt-dev-38/lib/python3.8/site-packages/pyproject_api/_backend.py True setuptools.build_meta
py38-docs: install_package_deps> python -I -m pip install 'numpy>=1.17.5' 'packaging>20.9' 'sympy>=1.5'
py38-docs: install_package> python -I -m pip install --force-reinstall --no-deps /Users/robcleme/dev/yt-project/unyt/.tox/.tmp/package/1/unyt-2.9.2.post290+g1029932.tar.gz
py38-docs: commands[0] /Users/robcleme/dev/yt-project/unyt/docs> make clean
py38-docs: failed with make is not allowed, use allowlist_externals to allow it
.pkg: _exit> python /Users/robcleme/.pyenv/versions/unyt-dev-38/lib/python3.8/site-packages/pyproject_api/_backend.py True setuptools.build_meta
  py38-docs: FAIL code 1 (44.81 seconds)
  evaluation failed :( (44.90 seconds)
```

neutrinoceros:
Okay so the CI part is now operational.
I would still need to think about how to rephrase the dev guide. I guess the simplest approach is to stop mentioning `pyenv` completely and only go pure tox. Does this sound acceptable to you @ngoldbaum ?

ngoldbaum:
No, please leave the pyenv instructions in. It’s the most straightforward cross-platform way I know of to simultaneously have multiple python versions available.

Why does the dev guide need to change at all? I thought tox-pyenv was only needed for github actions?

neutrinoceros:
> Why does the dev guide need to change at all? I thought tox-pyenv was only needed for github actions?

tox-pyenv is explicitly mentioned in the instructions as `pip install tox tox-pyenv`

ngoldbaum:
Hmm that’s interesting. I don’t think I wrote that? I don’t use it locally.

ngoldbaum:
Hmm I guess I did? I don’t actually use it though so I don’t think it’s necessary.

neutrinoceros:
Alright, great ! thank you for answer so quickly !

neutrinoceros:
tox 4.1.0 was just released, let's refresh CI to make sure it's still viable

PR #345, opened 2022-12-26
TST: cleanup redundant requirements in tox.ini

neutrinoceros:
`setuptool_scm` is only a dependency at buildtime and is already specified in `pyproject.toml` so we don't need it in `tox.ini`

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -23,7 +23,6 @@ deps =
     matplotlib!=3.5.0
     docutils
     dask[array,diagnostics]
-    setuptools_scm
 commands =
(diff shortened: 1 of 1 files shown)


neutrinoceros:
We could probably also save repeating runtime dependencies that are already specified in `pyproject.toml`, I'll give it a try

tox.ini
```diff
@@ -10,20 +10,16 @@ recreate = true
...
-    sympy
-    numpy
     h5py
     pint
     astropy
     coverage[toml]>=5.0
-    packaging
     pytest-cov
     pytest-doctestplus
     setuptools
```
neutrinoceros, line 19: Ah, I missed this line
```suggestion
```

PR #351, opened 2023-01-04
ENH: (NEP 18) test and implement combinatory functions

neutrinoceros:
Covers most of the functions last disscussed in https://github.com/yt-project/unyt/issues/289#issuecomment-1369294217

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -1,4 +1,5 @@
 import warnings
+from numbers import Number
 
 import numpy as np
 from packaging.version import Version
@@ -152,17 +153,19 @@ def histogram_bin_edges(a, *args, **kwargs):
     )
 
 
-def get_units(arrays):
+def get_units(objs):
     units = []
-    for sub in arrays:
+    for sub in objs:
         if isinstance(sub, np.ndarray):
             units.append(getattr(sub, "units", NULL_UNIT))
+        elif isinstance(sub, Number):
+            units.append(NULL_UNIT)
         else:
             units.extend(get_units(sub))
     return units
 
 
-def _validate_units_consistency(arrays):
+def _validate_units_consistency(objs):
     """
     Return unique units or raise UnitInconsistencyError if units are mixed.
     """
@@ -170,12 +173,24 @@ def _validate_units_consistency(arrays):
     # by using this as a guard clause in unyt_array.__array_function__
     # because it's already a necessary condition for numpy to use our
     # custom implementations
-    units = get_units(arrays)
+    units = get_units(objs)
     sunits = set(units)
     if len(sunits) == 1:
         return units[0]
     else:
         raise UnitInconsistencyError(*units)
+
+
+def _validate_units_consistency_v2(ref_units, *args) -> None:
+    """
+    raise UnitInconsistencyError if units are mixed
+    if all args are pure numbers, they are treated as having ref_units,
+    otherwise they are treated as dimensionless
+    """
+    if all(isinstance(_, Number) for _ in args):
+        return
+    else:
+        _validate_units_consistency((1 * ref_units, *args))
 
 
 @implements(np.concatenate)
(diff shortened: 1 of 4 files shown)


unyt/tests/test_array_functions.py
```diff
@@ -249,6 +238,12 @@ def get_wrapped_functions(*modules):
     return dict(sorted(wrapped_functions.items()))
 
 
+def assert_array_equal_units(a, b):
```
ngoldbaum, line 241: Why can't you use the version of this that's already in `test_unyt_array`? Totally fine to modify that one to work with ndarrays too.

unyt/tests/test_array_functions.py
```diff
@@ -1371,4 +1366,201 @@ def test_pad():
...
+    a = [1, 2, 3] * cm
+    res = np.extract(a > 1 * cm, a)
+    assert type(res) is unyt_array
+    assert res.units == cm
+
+
+def test_fill_diagonal_mixed_units():
+    a = np.zeros(9).reshape((3, 3)) * cm
+    with pytest.raises(UnitInconsistencyError):
+        np.fill_diagonal(a, 1)
```
ngoldbaum, line 1396: I prefer this (and other similar cases below) to not error.

unyt/_array_functions.py
```diff
@@ -714,3 +717,159 @@ def cumprod(a, *args, **kwargs):
...
+def setdiff1d(ar1, ar2, *args, **kwargs):
+    retu = _validate_units_consistency((ar1, ar2))
+    return (
+        np.setdiff1d._implementation(np.asarray(ar1), np.asarray(ar2), *args, **kwargs)
+        * retu
+    )
+
+
+@implements(np.where)
+def where(condition, x, y, *args, **kwargs):
```
ngoldbaum, line 834: `x` and `y` are optional, so this will break someone doing `np.where(some_unyt_array)`.

I haven't carefully gone through elsewhere in the numpy API to check for whether optional arguments are handled correctly, it would be good to double-check that we're not making similar errors elsewhere in our wrapping.

unyt/tests/test_array_functions.py
```diff
@@ -1371,4 +1366,201 @@ def test_pad():
...
+    y = [0, 0, 0] * cm
+    res = np.where(x > y, x, y)
+    assert type(res) is unyt_array
+    assert res.units == cm
+
+
+def test_clip_mixed_units():
+    a = [1, 2, 3, 4, 5, 6] * cm
+    with pytest.raises(UnitInconsistencyError):
+        np.clip(a, 3 * cm, 4)
```
ngoldbaum, line 1560: I agree this should error, but I kinda think `np.clip(a, 3, 4)` shouldn't error. The fact that it does now breaks existing unyt tests is another indicator that it shouldn't error, since it could just as easily break downstream user code.

unyt/tests/test_array_functions.py
```diff
@@ -1371,4 +1366,201 @@ def test_pad():
...
+    arr = np.arange(6, dtype=np.int_).reshape(2, 3) * cm
+    np.putmask(arr, arr > 2 * cm, np.zeros_like(arr))
+
+    assert type(arr) is unyt_array
+    assert arr.units == cm
+
+
+def test_searchsorted_mixed_units():
+    with pytest.raises(UnitInconsistencyError):
+        np.searchsorted([1, 2, 3, 4, 5] * cm, 3)
```
ngoldbaum, line 1485: Again, don't think this should error either.

ngoldbaum (commented):
I'm worried that some of the new errors that will get raised here will break downstream code in a somewhat annoying way. I would err on the side of only checking unit consistency if one of the arguments has a unit, otherwise if all arguments are numbers or ndarrays, assume that the arguments have units appropriate for the array.

unyt/_array_functions.py
neutrinoceros: Good call, I didn't realise they were optional.

> it would be good to double-check that we're not making similar errors elsewhere in our wrapping.

agreed, however I am currently nowhere near the point where I'd feel like revisiting all of them.

unyt/tests/test_array_functions.py
neutrinoceros: >  but I kinda think np.clip(a, 3, 4) shouldn't error

me too TBH, but in that case it seems to me that making `np.clip(a, 3 * cm, 4)` error will add more complexity than it is worth.

neutrinoceros:
> I'm worried that some of the new errors that will get raised here will break downstream code in a somewhat annoying way. I would err on the side of only checking unit consistency if one of the arguments has a unit, otherwise if all arguments are numbers or ndarrays, assume that the arguments have units appropriate for the array.

That is also how I think it _should_ work, and I agree that the fact that I had to change existing tests is indication that it's currently on the wrong track.
So to be clear in those cases I should treat pure numbers as if they had the appropriate units instead of treating them as dimensionless ?

ngoldbaum:
> So to be clear in those cases I should treat pure numbers as if they had the appropriate units instead of treating them as dimensionless?

Yes, but again only if all of the arguments are pure numbers. If one of them has units, I think it makes sense to do unit checking.

unyt/tests/test_array_functions.py
neutrinoceros: mostly because I expected it to live in a public API module and wanted to avoid importing from a test module.
Can I move the definition to `unyt/testing.py` ?

ngoldbaum: please do

unyt/tests/test_array_functions.py
neutrinoceros: nevermind it's actually quite simple

neutrinoceros:
@ngoldbaum I just pushed a changeset that I think should resolved on-going discussions, however I need to make another pass tomorrow to make sure I've updated all other functions to be in line with the behaviour we want.

neutrinoceros:
Alright this is good for review again. To make changes easier to review I just pushed incremental changes, but 
I would like my commits to get squashed before or on merge. I can do it with a rebase once we have converged on reviews, just please don't merge the branch in its current state.

unyt/testing.py
```diff
@@ -49,6 +51,12 @@ def assert_allclose_units(actual, desired, rtol=1e-7, atol=0, **kwargs):
         raise AssertionError
 
 
+def assert_array_equal_units(a, b):
```
ngoldbaum, line 54: Now that it's in the public API it needs a docstring

ngoldbaum:
Sorry for not getting to this last friday. Other than the missing docstring this looks good to me. Please go ahead and squash the commits as you'd like :)

neutrinoceros:
No worries ! I'll try to get this finished this evening.

neutrinoceros:
@ngoldbaum I force-pushed with a compact history. The change you requested in now in the second commit. It's now ready for merging.

Issue #296, opened 2022-10-15
Type checking unyt ?


neutrinoceros:
This is mostly a question to @ngoldbaum and @jzuhone:
How would you guys feel about progressively adding type hints and a type checking stage to CI ? To be clear I'm thinking about doing it at least partially myself, because numpy is almost 100% "typed" now and IMO it would make sense to follow their lead. This is a long term goal as this could be quite an undertaking (though maybe not !), so I wanted to get your sentiment on it first.

ngoldbaum:
I’m pretty ambivalent about it, I don’t think it’s worth putting a ton of effort in for marginal user benefit.

neutrinoceros:
The thing is that I really don't have a clear idea how much work it would take. Hypothetically, were the change actually not that big, would you consider it ?

ngoldbaum:
I don’t see how you could do it without a lot of code churn. If you’d like to try to convince me, go ahead, but you will need to convince me before I merge PRs adding type annotations.

neutrinoceros:
I think I need to convince myself first 😅
*if* I do, I may submit a proof of concept PR, but for now NEP 18 is much higher priority to me.

neutrinoceros:
I'd like to try doing this in progressive steps. That is, while I don't intend to add type hints everywhere in the code base in one go, I do see value in sprinkling them _as I'm working_ on various aspects of the code.
In order enable this, we first need to resolve (or explicitly ignore) all errors that `mypy unyt` can currently raise, and add type checking to CI. I have convinced myself this first step is a reasonable amount of work so I will open a PR to do just that. After that, we can add type hints wherever and whenever we want if it feels right.

ngoldbaum:
Like I said back in October, you’re going to need to convince me it’s worthwhile to add type annotations before I merge PRs adding them.

neutrinoceros:
Well an argument could be made that #355 is valuable in itself; because mypy in general will check consistency of one's code _and its dependencies_, by running type-checking in unyt's CI we can at a minimum guarantee that downstream code will not discover inconsistencies they cannot control within unyt. We've been working on very progressively adding type information in yt so I can testify that there's at least one code depending on unyt that does this.

Beyond that, I can think of a couple reasons why type annotations are useful to have for unyt's own good:

1) modern IDEs understand type hints and integrate them via hovering and auto-completion, which improves coding experience for anyone who relies on the annotated library.
[...]

PR #353, opened 2023-01-15
ENH: make dask a hard requirement for unyt.dask_array

neutrinoceros:
`unyt.dask_array` itself isn't loaded with `import unyt` so this doesn't make `dask` a hard dependency to the whole package, just the one module that is already unusable without it.

I find this simple change helps with boostrapping type-checking in unyt (see #296) because `mypy` is (rightfully) confused by conditional inheritance as is currently implemented in `unyt_dask_array`.

Namely, this resolves the following warnings from mypy
```
unyt/dask_array.py:194: error: Variable "unyt.dask_array._dask_Array" is not valid as a type  [valid-type]
unyt/dask_array.py:194: note: See https://mypy.readthedocs.io/en/stable/common_issues.html#variables-vs-type-aliases
unyt/dask_array.py: note: In class "unyt_dask_array":
unyt/dask_array.py:194: error: Invalid base class "_dask_Array"  [misc]
```

`unyt._on_demand_imports._dask` continues to be useful in `unyt_array.__array_ufunc__` so I'm not removing it.

### DIFF
--- a/unyt/dask_array.py
+++ b/unyt/dask_array.py
@@ -5,15 +5,24 @@ a dask array class (unyt_dask_array) and helper functions for unyt.
 
 """
 
+import sys
 from functools import wraps
 
 import numpy as np
-import pytest
 
 import unyt.array as ua
 
-pytest.importorskip("dask")
-del pytest
+if "pytest" in sys.modules:
+    # should only happen if pytest is installed *and* already imported,
+    # so we can skip collecting doctests from this module when dask isn't installed
+    # while avoiding making pytest itself a hard dependency to this module.
+    # This check is constructed to work with direct invocation (pytest unyt)
+    # as well as through python -m pytest
+    import pytest
+
+    pytest.importorskip("dask")
+    del pytest
+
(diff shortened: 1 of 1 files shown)


unyt/dask_array.py
```diff
@@ -8,20 +8,13 @@
 from functools import wraps
 
 import numpy as np
+import pytest
```
chrishavlin, line 11: this makes `pytest` a hard dependency for anyone attempting to use `unyt.dask_array`, doesn't it?

chrishavlin (commented):
simplifying the conditional import is great, but one question about the `pytest` import that you added (see comments).

unyt/dask_array.py
neutrinoceros: Good call, I actually missed that. I made some attempts to use protocols defined in pytest-doctest-plus but this is the only solution I found. I'll try to refine it.

neutrinoceros:
Thank you for pointing this out @chrishavlin ! should be good now

chrishavlin (approved):
Looks good!

PR #354, opened 2023-01-15
ENH: don't drop ImportError tracebacks for lazily-loaded optional dependencies

neutrinoceros:
equivalent to https://github.com/yt-project/yt/pull/4226
see https://github.com/yt-project/yt/issues/4224 for context

### DIFF
--- a/unyt/_on_demand_imports.py
+++ b/unyt/_on_demand_imports.py
@@ -2,7 +2,7 @@
 A set of convenient on-demand imports
 """
 
-
+import sys
 from functools import wraps
 from importlib.util import find_spec
 from typing import Type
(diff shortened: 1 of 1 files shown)


unyt/_on_demand_imports.py
```diff
@@ -40,7 +59,7 @@ def __init_subclass__(cls):
 
     def __new__(cls):
         if cls is OnDemand:
-            raise TypeError("The OnDemand base class cannot be instantiated.")
```
jzuhone, line 43: The old spelling is correct in American English, FWIW

neutrinoceros: I thought it was. Well, it only got there because I copy-pasted the code from yt again.

ngoldbaum:
This has a conflict. I'd also prefer it if you removed the types.

neutrinoceros:
rebased without the conflict, and typing-free

PR #366, opened 2023-02-06
DOC: sync HISTORY.rst from branch 2.9.x

neutrinoceros:


munkm:
Why was this submitted without a PR description and then merged without any feedback or review? What example does this set for the contributors we want to bring into the community?

PR #371, opened 2023-02-22
BUG: hotfix regression #369

### LINKED ISSUE #369: Zero comparison - breaking change?
I've been looking into `unyt` as a dependency of another package, and noticed that the change in behaviour in `2.9.4` from `2.9.2` (latest previous version tested _**Edit**: behaviour arises in v2.9.3, see below_) regarding zero comparison. This breaks a number of aspects of the  dependent code.

Illustrating the issue, here is the `2.9.4` behaviour:

```
>>> unyt.__version__
'v2.9.4'
>>> a = unyt.unyt_array(np.zeros(5, dtype=np.int8), dtype=np.int8)
>>> for i in a:
...     print(i == 0)
... 
False
False
False
False
False
>>> 
```

Then with `unyt` `2.9.2` (latest previous version tested):

```
>>> unyt.__version__
'v2.9.2'
>>> a = unyt.unyt_array(np.zeros(5, dtype=np.int8), dtype=np.int8)
>>> for i in a:
...     print(i == 0)
... 
True
True
True
True
True
>>> 
```

should this be regarded a breaking change, i.e. correspond to a new major version number?

neutrinoceros:
opening as a draft with just the test so we can see it fail first,
will close #369

### DIFF
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -2746,3 +2746,9 @@ def test_string_comparison():
     a = 1 * cm
     assert not (a == "hello")
     assert a != "hello"
+
+
+def test_int8_comparison():
+    # see regression https://github.com/yt-project/unyt/issues/369
+    a = unyt_array(np.zeros(5, dtype=np.int8))
+    assert all(e == 0 for e in a)


neutrinoceros:
So AFAICT the error was that `'b'` was incorrectly used as the `dtype.char` for `bool`, when it's really mapped to `int8` (instead, `np.dtype(bool).char == '?'`).
I haven't yet found a reliable source for the type -> char mapping. The closest thing I can find is this comment from numpy's source code, which contains the same mistake I did (mapping `bool` to `'b'`)

https://github.com/numpy/numpy/blob/ef336efec4c31da909f3091dce6f998dbe6e3861/numpy/core/numerictypes.py#L545-L555

So far, I don't know what `'a'` is supposed to be mapped to.

neutrinoceros:
Adding booleans to `DISALLOWED_DTYPE` breaks another test, and I think the idea is to forbid non-numeric types, and `bool` is arguably a numeric type in Python, so it probably doesn't belong here any way.

ngoldbaum:
Yeah, `?` is bool. 'a' is an alias for bytestring, so its `char` is `'S'`:

```
In [5]: np.dtype('a').char
Out[5]: 'S'
```

There's also `np.typecodes`, if that's helpful.

neutrinoceros:
Awesome, thank you !
IIRC `bytestring` is Python2 only, so we don't need to list it there, right ?

ngoldbaum:
No, numpy still supports bytestrings and will forever probably.

neutrinoceros:
Ok then I'll just fill in the comment and squash my history, then it'll be ready for review. I can do a quick bugfix release for it too if that's okay with you.

ngoldbaum:
> I can do a quick bugfix release for it too if that's okay with you.

Please! Thanks for handling this so quickly.

unyt/array.py
```diff
@@ -131,7 +131,14 @@
 
 NULL_UNIT = Unit()
 POWER_MAPPING = {multiply: lambda x: x, divide: lambda x: 2 - x}
-DISALLOWED_DTYPES = ("S", "U", "a", "O", "M", "m", "b")
+DISALLOWED_DTYPES = (
+    "S",  # string
+    "U",  # (unicode) bytes
+    "a",  # bytestring
```
ngoldbaum, line 137:
```suggestion
    "S",  # bytestring
    "a",  # bytestring
    "U",  # (unicode) bytes
```

Small fix, both "a" and "S" are bytestring.

PR #375, opened 2023-02-22
Add `__setitem__` implementation

### LINKED ISSUE #373: Unexpected behaviour when assigning array elements
* unyt version: 2.9.4 (but was always present)
* Python version: 3.10.6 (though not relevant)
* Operating System: irrelevant

### Description

No unit check is performed when assigning elements of an array. This is fairly counter-intuitive behaviour
and can likely lead to user errors downstream.

Here is an example (courtesy in part of @JBorrow):

```python
In [1]: import unyt

In [2]: unyt.__version__
Out[2]: 'v2.9.4'

In [3]: x = unyt.unyt_array([0, 2, 3], "cm")

In [4]: y = unyt.unyt_array([5, 6, 7], "kg")

In [5]: x[1] = 4

In [6]: x[2] = y[0]

In [7]: x
Out[7]: unyt_array([0, 4, 5], 'cm')
```

Lines 5 and 6 show two different problematic cases. Line 5 assigns a dimensionless quantity to a dimensionful array.  That may be OK but is ambiguous. Line 6 is especially problematic as we are now assigning completely invalid units to each other and the code is just happy to go along.

ngoldbaum:
Fixes #373 

There may be other ways to break things via `__setitem__` but this at least fixes the cases raised in the issue. If anyone spots additional things I should be checking, please let me know.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1726,10 +1726,8 @@ class unyt_array(np.ndarray):
         return ret
 
     def __setitem__(self, item, value):
-        value_u = getattr(value, "units", None)
-        if value_u is not None:
-            if value_u != self.units:
-                value = value.to(self.units)
+        if hasattr(value, "units") and value.units != self.units:
+            value = value.to(self.units)
         super().__setitem__(item, value)
 
     def __pow__(self, p, mod=None, /):
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1725,6 +1725,13 @@ class unyt_array(np.ndarray):
             ret.units = self.units
         return ret
 
+    def __setitem__(self, item, value):
+        value_u = getattr(value, "units", None)
+        if value_u is not None:
+            if value_u != self.units:
+                value = value.to(self.units)
+        super().__setitem__(item, value)
+
     def __pow__(self, p, mod=None, /):
         """
         Power function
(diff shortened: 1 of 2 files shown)


unyt/array.py
```diff
@@ -1725,6 +1725,13 @@ def __getitem__(self, item):
             ret.units = self.units
         return ret
 
+    def __setitem__(self, item, value):
+        value_u = getattr(value, "units", None)
+        if value_u is not None:
+            if value_u != self.units:
+                value = value.to(self.units)
```
neutrinoceros, line 1732: mostly stylistic, although the behaviour is ever so slightly different (when both `self` and `value` have `None` for a `units` attribute, which of course should never happen)
```suggestion
        if hasattr(value, "units") and value.units != self.units:
            value = value.to(self.units)
```

neutrinoceros: also, question: is this interoperable with `astropy.units` and `pint` ? (just curious)

neutrinoceros (approved):
one nit and a question, but otherwise LGTM

unyt/array.py
ngoldbaum: Astropy has `value.unit` instead of `value.units`, so that won't work. If I try to call `__setitem__` on a `unyt_array` with an astropy Quantity, I get a `TypeError` from astropy, but that's not new behavior with this PR.

For Pint I get a complicated traceback from Pint internals after calling `value.to(self.units)`.

I'm not overly upset over either situation.

ngoldbaum: I have an irrational hatred of `hasattr` but you're right that's simpler.

neutrinoceros (approved):

unyt/array.py
neutrinoceros: > I'm not overly upset over either situation.

me neither, just asking !

> I have an irrational hatred of hasattr

yeah, I get it. It can easily degenerate into terrible OOP.

PR #379, opened 2023-02-27
Fix Planck length tex_repr.

haymanpf:
Planck length units throw a TeX error when plotting in YT, which seems to be due to missing braces in the corresponding tex_repr.  This fixed the problem for me.


ngoldbaum:
Thanks for the catch! I pushed a commit to your branch which should fix the test your fix broke. I'll merge once I see the CI is green.

PR #398, opened 2023-03-28
ENH: (NEP 18) implement and test np.in1d and np.interp

neutrinoceros:
As discussed in #289

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -756,6 +756,12 @@ def isin(element, test_elements, *args, **kwargs):
     )
 
 
+@implements(np.in1d)
+def in1d(ar1, ar2, *args, **kwargs):
+    _validate_units_consistency((ar1, ar2))
+    return np.isin._implementation(np.asarray(ar1), np.asarray(ar2), *args, **kwargs)
+
+
 @implements(np.place)
 def place(arr, mask, vals, *args, **kwargs) -> None:
     _validate_units_consistency_v2(arr.units, vals)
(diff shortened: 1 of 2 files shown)


neutrinoceros:
actually, going into draft mode for a minute, while I try to squeeze `np.interp` in too

neutrinoceros:
did I forget to click the "draft" button ? Anyway, this is ready now

unyt/_array_functions.py
```diff
@@ -938,3 +944,13 @@ def tensordot(a, b, *args, **kwargs):
 def unwrap(p, *args, **kwargs):
     ret_units = p.units
     return np.unwrap._implementation(p.view(np.ndarray), *args, **kwargs) * ret_units
+
+
+@implements(np.interp)
+def interp(x, xp, fp, *args, **kwargs):
+    _validate_units_consistency((x, xp))
+    ret_units = getattr(fp, "units", 1)
```
ngoldbaum, line 952:
```suggestion
    ret_units = getattr(fp, "units", NULL_UNIT)
```

ngoldbaum: Doesn't really make much difference, the `1` just made me double-take reading this, `NULL_UNIT` makes the logic clearer.

neutrinoceros: I intentionally didn't use `get_units` (which is equivalent to your suggestion) because I wanted the output to match fp's type (so a pure ndarray wouldn't leak as a dimensionless unyt_array). I could add a comment to make that intention clearer, or I could just go with your suggestion and change the test too, no strong opinions actually. Your call !

ngoldbaum: Ah good call please add a comment

PR #399, opened 2023-03-28
ENH: (NEP 18) follow NEP 18 more closely for unsupported functions

neutrinoceros:
Taking care of `np._ix` and going down the rabbit hole in `astropy`, found about this rule in NEP 18 that unsupported functions should yield a `NotImplemented` sentinel value (which _then_ triggers a `TypeError`).

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -13,6 +13,32 @@ from unyt.exceptions import (
     UnytError,
 )
 
+# Functions for which passing units doesn't make sense
+# bail out with NotImplemented (escalated to TypeError by numpy)
+_UNSUPPORTED_FUNCTIONS = {
+    # Polynomials
+    np.poly,
+    np.polyadd,
+    np.polyder,
+    np.polydiv,
+    np.polyfit,
+    np.polyint,
+    np.polymul,
+    np.polysub,
+    np.polyval,
+    np.roots,
+    np.vander,
+    # datetime64 is not a sensible dtype for unyt_array
+    np.datetime_as_string,
+    np.busday_count,
+    np.busday_offset,
+    np.is_busday,
+    # not clear how to approach
+    np.piecewise,  # astropy.units doens't have a simple implementation either
+    np.packbits,
+    np.unpackbits,
+}
+
 _HANDLED_FUNCTIONS = {}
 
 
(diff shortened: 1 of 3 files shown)


neutrinoceros:
getting actual errors from docs examples. Switching to draft while I figure it out.

docs/usage.rst
```diff
@@ -1149,7 +1149,7 @@ this example:
   >>> from unyt import cm
   ...
   >>> data = [1, 2, 3]*cm
-  >>> np.save('my_data_cm.npy', data)
+  >>> np.save('my_data_cm.npy', data.d)
```
neutrinoceros, line 1152: Given this example exists mostly to show how unpractical `np.save` is with unyt_arrays, I this it's probably okay to add this small amount of additional complexity ?

unyt/_array_functions.py
```diff
@@ -13,6 +13,37 @@
...
+    np.datetime_as_string,
+    np.busday_count,
+    np.busday_offset,
+    np.is_busday,
+    # not clear how to approach
+    np.piecewise,  # astropy.units doens't have a simple implementation either
+    np.packbits,
+    np.unpackbits,
+    np.i0,
+    np.ix_,
```
ngoldbaum, line 44: Can we split this into functions that are truly unsupported and we want to raise errors and functions where we're fine with numpy ignoring the units?

`np.save` ignoring units might be fine for some existing use cases, especially if someone were following our docs which explicitly gave `np.save` as an example of something you can do with a `unyt_array`. I don't really see a reason to break that if it used to work, even if it might be slightly less confusing for future users. If we really want to break it I'd prefer if we did a deprecation cycle first.

In addition to `np.save`, `np.savez` and `np.savez_compressed`, I don't think it really hurts anything if `np.i0` ignored units (e.g. like `np.sin` does and will continue to do).

I think it's fine for all the other ones to error and it's unlikely that someone has used them with a `unyt_array` already, but if you want to go over the list and feel like it might make sense for some of them to work without units, feel free to depart from what I'm suggesting.

neutrinoceros: Sure, I can make this more subtle. Will edit in a minute.

neutrinoceros: done. I've also left `np._ix` out of the PR to avoid conflicts with #398

PR #404, opened 2023-04-01
DEPR: deprecate unyt.testing.assert_allclose_units

neutrinoceros:
Not ready yet, as I'm facing an unexpected challenge to get `np.testing.assert_allclose` to behave properly if passed two arrays with different units *and* different dimensions, which is something I missed in #200 when I originally planned out this deprecation.

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -483,10 +483,14 @@ def _array_comp_helper(a, b):
 
 @implements(np.isclose)
 def isclose(a, b, *args, **kwargs):
-    a, b = _array_comp_helper(a, b)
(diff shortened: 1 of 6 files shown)


neutrinoceros:
Taking a step back, this is probably doomed. However I still think it's possible to take advantage of recent changes to improve error messages in testing functions. I'll take a couple hours to sit on it but will probably just close this in favor of a different approach

neutrinoceros:
Making error message clearer is a much more substantial effort than I anticipated, mostly because of edge cases with temperature units, so I think I'll simply bail out on this for now and just cleanup the commented deprecation message.

PR #412, opened 2023-04-10
BUG: fix comparison for temperature arrays with different units

### LINKED ISSUE #407: BUG: temperature units comparison

neutrinoceros:
closes #407

### DIFF
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -806,7 +806,7 @@ def test_isclose(a, b, expected):
     np.testing.assert_array_equal(res, expected)
 
 
-def test_iclose_error():
+def test_isclose_error():
     x = [1, 2, 3] * cm
     y = [1, 2, 3] * g
     with pytest.raises(UnitConversionError):


unyt/_array_functions.py
```diff
@@ -469,10 +469,9 @@ def _array_comp_helper(a, b):
     au = getattr(a, "units", NULL_UNIT)
     bu = getattr(b, "units", NULL_UNIT)
     if bu != au and au != NULL_UNIT and bu != NULL_UNIT:
-        if (bu / au).is_dimensionless:
-            b = np.array(b) * (1 * bu).to(au)
-        else:
-            raise UnitConversionError(au, au.dimensions, bu, bu.dimensions)
+        if not bu.same_dimensions_as(au):
+            UnitConversionError(au, au.dimensions, bu, bu.dimensions)
```
ngoldbaum, line 473: this needs a `raise` statement

ngoldbaum: and i guess a test for the exception case :stuck_out_tongue:

neutrinoceros: How did I miss this ???

neutrinoceros: interestingly, this test already exists *and* already passes because the error is raised from the following line when conversion fails.
The test has a typo in its name though, so I'll fix it, and remove this redundant (and broken !) branch.

neutrinoceros:
pre-commit CI and tests both crash with a dubious `NameError` that I cannot reproduce locally... let's try rewriting the history of the branch to re-trigger everything 🤞🏻

PR #413, opened 2023-04-10
BUG: fix subtraction for different dimensionally equivalent units

### LINKED ISSUE #220: Subtracting temperatures gives degC rather than delta_degC
* unyt version:
'v2.8.0'
* Python version:
3.8.12
* Operating System:
Mac

### Description

Working with change in temperature. I would expect a difference between two temperatures to return a delta_degC rather than a degC. 

### What I Did

Actual
```
>>> import unyt as u
>>> delta = 2 * u.degC - 1 * u.degC 
>>> delta 
unyt_quantity(1, 'degC')
>>> delta.to('K')
unyt_quantity(274.15, 'K')
```

Expected

```
>>> import unyt as u
>>> delta = 2 * u.degC - 1 * u.degC 
>>> delta 
unyt_quantity(1, 'delta_degC')
>>> delta.to('K')
unyt_quantity(1, 'K')
```

neutrinoceros:
end goal: fix #220
This is very preliminary, as multiple tests are currently broken, but it's getting late, so I'm opening to get this out of my system for today.

### DIFF
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -2493,6 +2493,26 @@ def test_delta_degF():
     assert 2 * delta_degF == unyt_quantity(2, "delta_degF")
 
 
+@pytest.mark.parametrize(
+    ("u0", "u1", "uout"),
+    [
+        (K, K, K),
+        (R, R, R),
+        (degC, degC, delta_degC),
+        (degF, degF, delta_degF),
+        (degC, delta_degC, degC),
+        (delta_degC, degC, degC),
+        (degF, delta_degF, degF),
+        (delta_degF, degF, degF),
+    ],
+)
+def test_delta_temperature_diff(u0, u1, uout):
+    # using repr comparison because
+    # 1) we don't care that Unit instances might not be the same
+    # 2) some temperature units will compare as equal even when they are not identical (e.g. K and delta_degC)
+    assert repr((2 * u0 - 1 * u1).units) == repr(uout)
+
+
 def test_mil():
     assert_allclose_units(unyt_quantity(1, "mil"), unyt_quantity(0.001, "inch"))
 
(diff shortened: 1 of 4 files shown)


ngoldbaum:
Let me know if you'd like a hand with this.

neutrinoceros:
Sure ! I'm actually having a hard time understanding how my changes affect tests `unyt/tests/test_unyt_array.py`

ngoldbaum:
> I'm actually having a hard time understanding how my changes affect tests `unyt/tests/test_unyt_array.py`

See [d708386](https://github.com/yt-project/unyt/pull/413/commits/d708386c3bd643ba2bff1429a16f9fa926aa9ccc)

neutrinoceros:
awesome, well played ! The one remaining failure (in docs) touches a case that I'm not sure how to treat. Should `1*delta_degC - 1*degC` remain a valid operation ?

ngoldbaum:
It looks like Pint returns data in units of `degC`, so I think we should probably follow that. I guess one way to justify it is:

```
1*delta_degC - 1*degC = -1 * (1*degC - 1*delta_degC)
```

So it makes sense for both operations to return results in the same units.

It also means we're not raising a new error without a deprecation, which is nice.

unyt/tests/test_unyt_array.py
```diff
@@ -2493,6 +2493,22 @@ def test_delta_degF():
     assert 2 * delta_degF == unyt_quantity(2, "delta_degF")
 
 
+@pytest.mark.parametrize(
+    ("u0", "u1", "uout"),
+    [
+        (degC, degC, delta_degC),
+        (K, K, delta_degC),
```
ngoldbaum, line 2500: Sorry to be a pain, but Kelvin minus Kelvin should result in Kelvin, similarly Rankine minus Rankine. Maybe the checks in `_difference_units` should check `base_offset`?

neutrinoceros: good call. I updated the test and the implementation

PR #416, opened 2023-04-13
ENH: make all custom exceptions derive from UnytError. Document it

neutrinoceros:
Resolves another part of #346. What's missing is the documentation (other than docstring).

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -1303,7 +1303,7 @@ It is possible to catch any exceptions from unyt as
   ... except UnytError:
   ...     pass
 
-However, it is in general advised to only catch specifi exceptions types that
+However, it is in general advised to only catch specific exceptions types that
 are known-possible outcomes. All custom exceptions types live in the
 :mod:`unyt.exceptions` module and may be imported from there.
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -1286,6 +1286,28 @@ As for HDF5 data, the unit registry associated with the unit object is saved to
 the pickle. If you have custom units defined, the reloaded data will know about
 your custom unit and be able to convert data to and from the custom unit.
 
+Handling errors from :mod:`unyt`
+--------------------------------
+
+:mod:`unyt` sometimes raises exceptions with unique exception types, e.g., to signal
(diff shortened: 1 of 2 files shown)


unyt/exceptions.py
```diff
@@ -7,13 +7,31 @@
...
+
+    However, it is generally recommended to only capture exceptions as
+    specific as possible.
+    """
+
     pass
 
 
-class UnitOperationError(ValueError):
+class UnitOperationError(UnytError, ValueError):
```
neutrinoceros, line 34: disclaimer: using multiple inheritance is explicitly recommended against in the documentation https://docs.python.org/3/library/exceptions.html#inheriting-from-built-in-exceptions

So I'm kinda hoping that the fact `UnytError` doesn't hold any specialisation is a safe case.

neutrinoceros:
I added documentation under the "Integrating [unyt](https://unyt.readthedocs.io/en/stable/modules/unyt.html#module-unyt) Into a Python Library" section. Maybe it'd be a better fit in a different section, but that's what felt most natural to me. Feedback welcome !

docs/usage.rst
```diff
@@ -1286,6 +1286,28 @@ As for HDF5 data, the unit registry associated with the unit object is saved to
...
+  >>> from unyt import cm, s
+  >>> from unyt.exceptions import UnytError
+  >>> a = 1 * cm
+  >>> b = 1 / s
+  >>> try:
+  ...     a + b
+  ... except UnytError:
+  ...     pass
+
+However, it is in general advised to only catch specifi exceptions types that
```
ngoldbaum, line 1306:
```suggestion
However, it is in general advised to only catch specific exceptions types that
```

unyt/exceptions.py
ngoldbaum: I wish that comment in the python docs had more detail about how this might fail. Let's see if we get any complaints from users after doing a release.

ngoldbaum (commented):
Just one small typo

unyt/exceptions.py
neutrinoceros: agreed

PR #420, opened 2023-05-10
Add tests to improve coverage

ngoldbaum:
There are a bunch of new code paths that don't have tests, this adds some tests to cover them. It also deletes some old code paths that are no longer getting test coverage.

More explanation for the non-test changes:

* I noticed there was some inconsistency in which errors are raised by the `__array_function__` wrappers, so I decided to make all the errors raised there `InvalidUnitOperation`. This seems clearest to me.

* The `_sanitize_range` wrapper wasn't handling cases where the range is all unitless scalars, this adds support for that case.

* The tox coverage reports were broken so I fixed that.

* The custom code in the `except` block in `__eq__` and `__ne__` was never getting called so I deleted the `__eq__` and `__ne__` implementations.

* `clip` is now handled by the `__array_function__` wrapper, so I deleted the special handling for it in `__array_ufunc__`.

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -9,7 +9,6 @@ from unyt.dimensions import temperature
 from unyt.exceptions import (
     InvalidUnitOperation,
     UnitInconsistencyError,
-    UnytError,
 )
 
 # Functions for which passing units doesn't make sense
@@ -141,11 +140,17 @@ def _sanitize_range(_range, units):
     for i in range(ndim):
         ilim = _range[2 * i : 2 * (i + 1)]
         imin, imax = ilim
-        if not (hasattr(imin, "units") and hasattr(imax, "units")):
-            raise TypeError(
-                f"Elements of range must both have a 'units' attribute. Got {_range}"
+        himin = hasattr(imin, "units")
+        himax = hasattr(imax, "units")
+        if himin != himax:
+            raise InvalidUnitOperation(
+                "Elements of range must both have a 'units' attribute if one does. "
+                f"Got {_range}"
             )
-        new_range[i] = imin.to(units[i]).value, imax.to(units[i]).value
+        elif himin:
+            new_range[i] = imin.to(units[i]).value, imax.to(units[i]).value
+        else:
+            new_range[i] = imin, imax
     return new_range.squeeze()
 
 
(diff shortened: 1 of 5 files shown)


unyt/_array_functions.py
```diff
@@ -926,11 +931,9 @@ def einsum(subscripts, *operands, out=None, **kwargs):
     ret_units = _validate_units_consistency(operands)
 
     if out is not None:
-        out_view = out.view(np.ndarray)
+        res = np.einsum._implementation(subscripts, *operands, out=out.view(np.ndarray))
```
neutrinoceros, line 934: this patch made me realise that this wrapper is swallowing kwargs other than `out` silently

unyt/array.py
```diff
@@ -1765,18 +1764,6 @@ def __pow__(self, p, mod=None, /):
         else:
             return super().__pow__(p, mod)
 
-    def __eq__(self, other):
-        try:
-            return super().__eq__(other)
-        except (IterableUnitCoercionError, UnitOperationError):
-            return np.zeros(self.shape, dtype="bool")
```
neutrinoceros, line 1772: have you tried running tests + coverage with numpy's bleeding edge version ? this code was supposed to support numpy 1.25 (yet to be released) (see https://github.com/yt-project/unyt/pull/333)

unyt/array.py
```diff
@@ -1962,26 +1949,10 @@ def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
...
-                ret_class = type(inputs[0])
-                mul = 1
-            else:
-                raise RuntimeError(
-                    "Support for the %s ufunc with %i inputs has not been "
-                    "added to unyt_array." % (str(ufunc), len(inputs))
-                )
+            raise RuntimeError(
+                "Support for the %s ufunc with %i inputs has not been "
+                "added to unyt_array." % (str(ufunc), len(inputs))
```
neutrinoceros, line 1954: while we're at it
```suggestion
                f"Support for the {ufunc} ufunc with {len(inputs)} "
                "inputs has not been added to unyt_array."
```

unyt/tests/test_array_functions.py
```diff
@@ -408,6 +411,18 @@ def test_histogram():
     assert type(counts) is np.ndarray
     assert bins.units == arr.units
 
+    counts, bins = np.histogram(
```
neutrinoceros, line 414: one general comment on this test module: I would prefer that new, independent test cases be added as separate test functions

neutrinoceros (commented):
Seems cool overall, thank you for doing this. I have some doubts about removing custom implementations for `__ne__` and `__eq__` on the basis that they are not covered by tests because they are meant to be hit only with currently unreleased versions of numpy.

PR #421, opened 2023-05-12
TST: add test for new regression with sympy 1.12

neutrinoceros:
There's a new incompatibility where a Unit object that was pickled with sympy 1.8 doesn't compare correctly with the same object generated with sympy 1.12. Unfortunately this case wasn't sufficiently tested so it was discovered _after_ sympy 1.12 was released, within yt test infrastructure.

This adds completes the regression test.

I'll report the issue upstream later today. I've identified the first broken commit as https://github.com/sympy/sympy/commit/40f0f6cfa2af216247476d3f416c715fefde6335

### DIFF
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -19,7 +19,6 @@ from pathlib import Path
 
 import numpy as np
 import pytest
-import sympy
 from numpy import array
 from numpy.testing import (
     assert_almost_equal,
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Depending on how the upstream issue evolve, we might not want to merge this, so I'm keeping it as a draft for now.

jzuhone:
@neutrinoceros what is the status of this?

neutrinoceros:
@jzuhone the problem is still here (the patch upstream was not merged yet), so I think this PR is still relevant

PR #426, opened 2023-06-12
TST: minimize build time in tox runs

neutrinoceros:
Just learned that tox 4 allows universal wheels to be built once and reused across multiple envs, reducing the build time to a minimum.
https://tox.wiki/en/latest/upgrading.html#universal-wheels

### DIFF
--- a/tox.ini
+++ b/tox.ini
@@ -98,4 +98,4 @@ commands =
 skip_install = true
 depends = py{38,39,310,311}
 deps =
-    coverage
+    coverage[toml]
--- a/tox.ini
+++ b/tox.ini
@@ -10,6 +10,8 @@ python =
     3.11: py311
 
 [testenv]
+package = wheel
+wheel_build_env = .pkg
 setenv =
     PYTHONPATH = {toxinidir}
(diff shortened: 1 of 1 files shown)


neutrinoceros:
This seems to save about 40% overhead on the Python 3.8 job in CI, but that's it, because it's the only job with multiple tasks to run. Probably still worth it.

Also fixed an issue with `testenv:end` where coverage was run without toml support and crashed.

neutrinoceros:
Apparently `setup-python` now supports setting multiple python versions in a single job
https://github.com/actions/setup-python/blob/main/docs/advanced-usage.md#specifying-multiple-pythonpypy-version
Switching to draft while I attempt to leverage this here

jzuhone:
Is this still worth it, considering we're probably going to drop Python 3.8 soon?

neutrinoceros:
I think so. The workload currently carried by Python 3.8 will eschew to 3.9 and we'll also add 3.12 when enough of our test dependencies support it, so I don't think this cycle is going to change much in itself.

PR #429, opened 2023-06-20
PERF: lazily define physical constants aliases

neutrinoceros:
This reduces by half the import penalty from unyt itself. Meaning, following `import numpy; import sympy; import sympy.tensor.tensor`, `import unyt` costs 250ms on my machine, and down to 120ms with this branch.

The idea is to avoid generating hundreds of rarely used quantities (based on how often they appear in tests, or yt) on import.
This comes a the cost of a (small ?) breaking changea:
`from unyt import *` won't import lazily generated aliases to physical constants (like `mass_sun`, `c_mks` ...). They are now only accessible if imported explicitly.

It may be possible to alleviate this by defining `__all__` module attributes. However I don't see a way to do it dynamically (cases where `UnitsNotReducible` is raised should not be included), and I'm not sure doing it statically is worth it.

Note that I'm keeping `unyt.unit_systems.add_constants` even if not used internally any more because it is still used in yt, but it should be possible to avoid it there too and deprecate it here.

The cost in breaking change in my eye small and worth the gain, but it's still breaking, so I'm proposing this for inclusion in the 3.0 release.

### DIFF
--- a/unyt/__init__.py
+++ b/unyt/__init__.py
@@ -62,21 +62,14 @@ from unyt.unit_systems import UnitSystem  # NOQA: F401
 from ._version import __version__
 
 
-# function to only import quantities into this namespace
-# we go through the trouble of doing this instead of "import *"
-# to avoid including extraneous variables (e.g. floating point
-# constants used to *construct* a physical constant) in this namespace
-def import_units(module, namespace):
-    """Import Unit objects from a module into a namespace"""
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Upon reflexion I am less and less sure this is worth the complexity and possible backward-compatibility breakage. I'll just close it as it also doesn't seem to raise interest.

PR #435, opened 2023-08-04
Allow dimension-checking of multiple return values

db434:
Extend the `@returns` decorator so it can handle functions with multiple return values.

Example:
```python
>>> @returns(length, length/time**2)
... def f(a, v):
...     return a * v, v / a
...
>>> res = f(a= 2 * u.s, v = 3 * u.m/u.s)
>>> print(*res)
6 m 1.5 m/s**2
```

I think I've followed everything in [CONTRIBUTING](https://github.com/yt-project/unyt/blob/main/CONTRIBUTING.rst), but please let me know if I've missed anything.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -329,6 +329,17 @@ using the :meth:`@accepts <unyt.dimensions.accepts>` and :meth:`@returns <unyt.d
   >>> print(res)
   6 m
 
+:meth:`@accepts <unyt.dimensions.accepts>` can specify the dimensions of any subset of inputs and :meth:`@returns <unyt.dimensions.returns>` must always describe all outputs.
+
+  >>> @returns(length, length/time**2)
+  >>> @accepts(v=length/time)
+  ... def bar(a, v):
+  ...     return a * v, v / a
+  ...
+  >>> res = bar(a= 2 * u.s, v = 3 * u.m/u.s)
+  >>> print(*res)
+  6 m 1.5 m/s**2
+
 .. note::
    Using these decorators may incur some performance overhead, especially for small arrays.
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -325,7 +325,7 @@ using the :meth:`@accepts <unyt.dimensions.accepts>` and :meth:`@returns <unyt.d
   ... def foo(a, v):
   ...     return a * v
   ...
-  >>> res = foo(a= 2 * u.s, v = 3 * u.m/u.s)
+  >>> res = foo(a=2*u.s, v=3*u.m/u.s)
   >>> print(res)
   6 m
 
@@ -336,7 +336,7 @@ using the :meth:`@accepts <unyt.dimensions.accepts>` and :meth:`@returns <unyt.d
   ... def bar(a, v):
   ...     return a * v, v / a
   ...
-  >>> res = bar(a= 2 * u.s, v = 3 * u.m/u.s)
+  >>> res = bar(a=2*u.s, v=3*u.m/u.s)
   >>> print(*res)
   6 m 1.5 m/s**2
 
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -332,7 +332,7 @@ using the :meth:`@accepts <unyt.dimensions.accepts>` and :meth:`@returns <unyt.d
 :meth:`@accepts <unyt.dimensions.accepts>` can specify the dimensions of any subset of inputs and :meth:`@returns <unyt.dimensions.returns>` must always describe all outputs.
 
   >>> @returns(length, length/time**2)
-  >>> @accepts(v=length/time)
+  ... @accepts(v=length/time)
   ... def bar(a, v):
   ...     return a * v, v / a
   ...
--- a/unyt/dimensions.py
+++ b/unyt/dimensions.py
@@ -359,7 +359,7 @@ def returns(*r_unit):
             if isinstance(results, tuple):
                 result_tuple = results
             else:
-                result_tuple = results,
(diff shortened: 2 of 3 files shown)


neutrinoceros:
This seems like a good idea ! I don't have time to review in detail *right now* but I'll be back very shortly. In the mean time, I suggest updating this decorator's documentation to reflect the new possibility. Thanks for your work !

db434:
Thanks for taking a look.

> In the mean time, I suggest updating this decorator's documentation to reflect the new possibility.

I think I've done this already - is there anywhere else that you would suggest changing?

neutrinoceros:
Thank you for following through ! automated testing seems to have caught some syntax error in the docs. I suggest you try to run `tox` locally to make sure everything is okay, because currently we need to manually approve your tests to run every time you push (this is a security measure against bad agents, it only applies to new contributors !).

db434:
Sorry for this CI spam - I've struggled to get all of the tests running locally. Hopefully this is the last time..!

docs/usage.rst
```diff
@@ -329,6 +329,17 @@ using the :meth:`@accepts <unyt.dimensions.accepts>` and :meth:`@returns <unyt.d
...
   6 m
 
+:meth:`@accepts <unyt.dimensions.accepts>` can specify the dimensions of any subset of inputs and :meth:`@returns <unyt.dimensions.returns>` must always describe all outputs.
+
+  >>> @returns(length, length/time**2)
+  ... @accepts(v=length/time)
+  ... def bar(a, v):
+  ...     return a * v, v / a
+  ...
+  >>> res = bar(a= 2 * u.s, v = 3 * u.m/u.s)
```
neutrinoceros, line 339: formatting
```suggestion
  >>> res = bar(a=2*u.s, v=3*u.m/u.s)
```

unyt/tests/test_unyt_testing.py
```diff
@@ -40,3 +42,67 @@ def test_atol_conversion_error():
...
     with pytest.raises(AssertionError):
         assert_allclose_units(a1, a2, atol=unyt_quantity(0.0, "kg"))
+
+
+def test_accepts():
+    @accepts(a=time, v=length / time)
+    def foo(a, v):
+        return a * v
+
+    _ = foo(a=2 * second, v=3 * meter / second)
```
neutrinoceros, line 52: If we're not going to use the return value, might as well not capture it (this comment applies to a handful more lines)
```suggestion
    foo(a=2 * second, v=3 * meter / second)
```

unyt/dimensions.py
```diff
@@ -290,14 +290,14 @@ def new_f(*args, **kwargs):
     return check_accepts
 
 
-def returns(r_unit):
+def returns(*r_unit):
```
neutrinoceros, line 293: this is a breaking API change to any user code that would use the argument as keyword
```python
@returns(r_unit=time)
def foo():
    ...
```
It is however possible to keep it backward compatible for now and deprecate this cumbersome and undocumented use case, like so:
```suggestion
def returns(*r_units, r_unit):
```
and with internal modifications to the function to handle all special cases where `r_unit` is provide:
  - raise an error if `r_units` is not empty
  - raise a warning that passing `r_unit` as keyword is deprecated, and assign`r_units = (r_unit,)`

neutrinoceros (commented):
This looks great ! thank you so much for working on this and adding tests ! I found a subtle way your change can break	backward compatibility for some users, and provided some indications on how to avoid that. Don't hesitate if you have any questions about that

unyt/dimensions.py
db434: Good catch! Any better now?

unyt/dimensions.py
```diff
@@ -318,9 +321,30 @@ def returns(r_unit):
...
+            raise ValueError(
+                "Cannot specify `r_unit` and other return values simultaneously"
+            )
+        else:
+            warnings.warn(
+                "Use of the @returns(r_unit=...) syntax is deprecated. "
+                "Please use @returns(...) instead.",
+                category=DeprecationWarning,
+                stacklevel=2,
+            )
```
neutrinoceros, line 345: we have a thin wrapper function for this in `unyt._deprecation` that I advise you use instead of a raw `DeprecationWarning`. I should have remembered and told you that earlier, sorry !

unyt/dimensions.py
neutrinoceros: yes, you're doing great ! And well done on the default value for the deprecated argument, I see now that I forgot to mention it.

unyt/dimensions.py
db434: Done. I'm not sure which version of unyt this will be deprecated in, so I went with 3.0 to match some other warnings I saw. Please let me know if you'd like that changed (or feel free to change it without me).

neutrinoceros (approved):
Awesome job ! The only missing part I can think of would be a couple tests that the deprecated API behave as expected, but I've tested it manually and I'd be okay to take the PR as is and add the tests later. I'll defer merging for a week or two so as to give other maintainers a chance to weigh in.

unyt/dimensions.py
neutrinoceros: 3.0 is indeed the next planned version, you're all good

db434:
I believe those tests are [here](https://github.com/db434/unyt/blob/a7ed932cc42faa1a1bbae8167c37a65e6303cacc/unyt/tests/test_unyt_testing.py#L88-L102). Is there anything else you're looking for?

Thanks for your patience with this!

jzuhone (approved):

PR #441, opened 2023-08-23
Various new non-metric (and one metric) units

jzuhone:
This PR adds a number of new non-metric and one new metric unit to the unit lookup table. Inspired originally by @ring-michael


ngoldbaum:
LGTM. If you can, it would help future readers to add comments with e.g. links to where the exact numerical constants are coming from.

unyt/_unit_lookup_table.py
```diff
@@ -186,7 +207,11 @@
         ("atm", (pascal_per_atm, dimensions.pressure, 0.0, r"\rm{atm}", False)),
         ("hp", (watt_per_horsepower, dimensions.power, 0.0, r"\rm{hp}", False)),
         ("oz", (kg_per_pound / 16.0, dimensions.mass, 0.0, r"\rm{oz}", False)),
-        ("ton", (kg_per_pound * 2000.0, dimensions.mass, 0.0, r"\rm{ton}", False)),
+        ("ton", (kg_per_pound * 2000.0, dimensions.mass, 0.0, r"\rm{US ton}", False)),
```
neutrinoceros, line 210:
```suggestion
        ("ton_US", (kg_per_pound * 2000.0, dimensions.mass, 0.0, r"\rm{US ton}", False)),
```

neutrinoceros (commented):
other than the missing suffix I found, LGTM !

unyt/_unit_lookup_table.py
jzuhone: So `ton` is already here and we weren't thinking about US vs UK when we implemented it (mistake). So I'm not sure how to handle it without breaking back-compat, other than making `ton_US` the unit and `ton` the alias (which I guess works).

neutrinoceros: Ah, right, I didn't realise it was already implemented ! nevermind then

neutrinoceros (approved):

PR #442, opened 2023-08-27
BUG: fix an incompatibility with numpy 2.0 (np.asfarray is removed)

neutrinoceros:
Fix a new incompatibility with numpy's dev branch (see https://github.com/numpy/numpy/pull/24376)

This bug only affects unyt 3.0, and the fix doesn't need a backport.


neutrinoceros:
@jzuhone can we merge this ?

jzuhone (approved):

PR #443, opened 2023-09-04
TST: fix testability against numpy 2.0 dev (handle numpy 1.25 deprecations)

neutrinoceros:
https://github.com/numpy/numpy/pull/24477 broke CI by removing `np.round_`, but in fact we hadn't systematically marked deprecations from numpy 1.25 so this fixes that.


neutrinoceros:
@jzuhone can we merge this ?

PR #445, opened 2023-09-06
BUG: fix an incompatibility with numpy 2.0 (np.trapz and np.in2d are deprecated)

neutrinoceros:
Following https://github.com/numpy/numpy/pull/24445

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -441,22 +441,6 @@ def fft_ifftshift(x, *args, **kwargs):
     )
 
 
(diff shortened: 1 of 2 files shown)


neutrinoceros:
I need to figure out wether `scipy.interpolate.trapezoid` can still be wrapped following NEP 18 or not

neutrinoceros:
> I need to figure out wether scipy.interpolate.trapezoid can still be wrapped following NEP 18 or not

it does not. While slightly annoying, this means the present PR is ready.

PR #448, opened 2023-09-25
TST: unlock numpy 2.0 on bleeding-edge CI

neutrinoceros:
At the moment, bleeding edge CI picks up matplotlib nightlies, but this results in downgrading numpy to a stable release (see https://github.com/matplotlib/matplotlib/issues/26847#issuecomment-1729927820)

matplotlib is not a hard dependency to unyt, and we're just using it to test the stability of integrations, but numpy is an absolute core dependency, so I think there's more value in trying to catch incompatibilities with it than with matplotlib.
This patch proposes to limit nightlies tests to core dependencies.


neutrinoceros:
@jzuhone for review

PR #455, opened 2023-10-18
BUG: fix future incompatibility with numpy 2.0 (numpy.core isn't public API)

neutrinoceros:
fix an on-going failure in bleeding-edge CI

Context:
https://github.com/numpy/numpy/pull/24634

The comment I'm removing is from https://github.com/yt-project/unyt/pull/204. Local testing suggests that it's not necessary anymore (tests introduced in the same PR still pass), I suspect this is due to bumping our minimal requirements for numpy since that PR.

Opening as a draft for validation against the full CI. If all goes well, I'll try to remove `unyt_array.__pow__` completely

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -102,7 +102,6 @@ from numpy import (
     true_divide,
     trunc,
 )
-from numpy.core.umath import _ones_like
 from sympy import Rational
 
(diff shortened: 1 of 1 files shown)


neutrinoceros:
Removing `unyt_array.__pow__` doesn't work, but all tests pass in the current state so this is ready for review

PR #456, opened 2023-10-26
TST: adapt a test to numpy 2.0 (NEP 50)

neutrinoceros:
This fixes tests with bleeding-edge numpy. I'm relatively confident this is the correct fix, but I want to come back to it later when I have more time to double check, so, opening as a draft for now.


neutrinoceros:
Now that I've had time to actually read NEP 50, I'm confident this is the correct approach. Opening for review.

neutrinoceros:
@jzuhone can we merge this ?

PR #457, opened 2023-10-27
Documentation for v3.0.0

jzuhone:
This is a documentation update for version 3.0.0. I added entries to `HISTORY` and `AUTHORS`, as well as fix some rendering issues with the deploying docs.

### DIFF
--- a/HISTORY.rst
+++ b/HISTORY.rst
@@ -2,6 +2,70 @@
 History
 =======
 
+3.0.0 (2023-10-27)
+------------------
+
+This new major release of ``unyt`` fixes a number of issues and adds new units.
+Major contributions include:
+
+* A number of new metric and non-metric units have been added
+  (`PR #441 <https://github.com/yt-project/unyt/pull/442>`_). Thank you
+  to John ZuHone (@jzuhone on GitHub) for the contribution.
+
+* The :ref:```@returns`` decorator<Checking Units>` now allows dimension-checking
+  of multiple return values (`PR #435 <https://github.com/yt-project/unyt/pull/435>`_).
+  Thank you to Daniel Bates (@db434 on GitHub) for the contribution.
+
+* A fix for for the LaTeX representation of Planck units
+  (`PR #379 <https://github.com/yt-project/unyt/pull/379>`_). Thank you to
+  Peter Hayman (@haymanpf on GitHub) for the contribution.
+
+* A fix for a bug that prevented the conversion of dimensionless arrays
+  to their corresponding `AstroPy Quantities <https://docs.astropy.org/en/stable/units/>`_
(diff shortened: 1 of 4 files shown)


matthewturk (approved):

HISTORY.rst
```diff
@@ -2,6 +2,72 @@
...
+  - `PR #434 <https://github.com/yt-project/unyt/pull/434>`_.
+  - `PR #442 <https://github.com/yt-project/unyt/pull/442>`_.
+  - `PR #443 <https://github.com/yt-project/unyt/pull/443>`_.
+  - `PR #445 <https://github.com/yt-project/unyt/pull/445>`_.
+  - `PR #448 <https://github.com/yt-project/unyt/pull/448>`_.
+  - `PR #455 <https://github.com/yt-project/unyt/pull/455>`_.
+  - `PR #456 <https://github.com/yt-project/unyt/pull/456>`_.
+
+* A number of PRs to support
+  `NEP 18 <https://numpy.org/neps/nep-0018-array-function-protocol.html>`_,
```
neutrinoceros, line 58: I think this list is incomplete, though it doesn't need to be: it should suffice to rephrase it 

```suggestion
* A number of PRs to support
  `NEP 18 <https://numpy.org/neps/nep-0018-array-function-protocol.html>`_, including the following
```

HISTORY.rst
```diff
@@ -2,6 +2,72 @@
...
+  Peter Hayman (@haymanpf on GitHub) for the contribution.
+
+* A fix for a bug that prevented the conversion of dimensionless arrays
+  to their corresponding `AstroPy Quantities <https://docs.astropy.org/en/stable/units/>`_
+  (`PR #437 <https://github.com/yt-project/unyt/pull/437>`_). Thank you to
+  Clément Robert (@neutrinoceros on GitHub) for the contribution.
+
+* A memory-management optimization obtained by avoiding unnecessary copies
+  in ``numpy.histogram*`` functions (`PR #425 <https://github.com/yt-project/unyt/pull/425>`_).
+  Thank you to Clément Robert (@neutrinoceros on GitHub) for the contribution.
```
neutrinoceros, line 30: This is an optimisation for NEP-18 code, so it's not a change wrt v2.9.5, I would move this to the NEP-18 section

neutrinoceros (commented):
Thank you for doing this John ! I think the implementation of NEP 18 should be much higher in the list of changes, since it's the reason why we are going for 3.0 instead of 2.10 : NEP 18 is an *improvement* in that it makes the behaviour of many numpy functions more intuitive (as well as more aligned with `astropy.units`, which implemented it first), but it's nonetheless a technically *breaking* change.

Issue #440, opened 2023-08-23
`unyt.amu` is a quantity, but `amu` is a built-in unit name as well.


chrisjonesBSU:
* unyt version: 2.9.5
* Python version: 3.11.4
* Operating System: MacOS (Intel)

### Description

I have a couple questions about the behavior of the `amu` unit.

### What I Did

First question: Why is the data type of amu different than other units like g, kg, g/mol, mm etc?

```
>>> import unyt as u

>>> type(u.amu)
<class 'unyt.array.unyt_quantity'>

>>> type(u.mg)
<class 'unyt.unit_object.Unit'>

>>> type(u.g/u.mol)
<class 'unyt.unit_object.Unit'>

>>> type(u.mm)
<class 'unyt.unit_object.Unit'>
```

Second question: When I try to create a new quantity with units of amu, the behavior is different than what I would expect. I get a unyt quantity in `kg` instead of `amu`

```
>>> 2.3 * u.amu
unyt_quantity(3.81923952e-27, 'kg')

>>> 2.3 * u.g
unyt_quantity(2.3, 'g')

>>> 2.3 * u.g/u.mol
unyt_quantity(2.3, 'g/mol')
```

I would expect that `2.3 * u.amu` would return `unyt_quantity(2.3, 'amu')`

It seems to work if I use `.to("amu")`. For example:

```
>>> 2.3 * u.amu.to("amu")
unyt_quantity(2.3, 'amu')
```

Is there something I'm missing about how Unyt handles amu compared to other mass units? 

Thanks!

ngoldbaum:
Hmm, this is an oversight in the library. 

The top-level `unyt` namespace is populated with both `Unit` objects, as you saw, and `unyt_quantity` objects, which are the unyt equivalent of numpy scalars - a single numeric value that has a unit attached. Things like `unyt.msun` are quantities, but `unyt.gram` are units.

I guess at some point AMU was defined only as a quantity, then subsequently it got added as a unit name as well and no one removed it as a quantity, so it shows up in the top-level namespace as a quantity. We should probably only have it show up as a unit, I think that's unambiguously more useful and should be backward compatible, but whoever does that should think carefully if that really is the case.

We should also probably add a check to make sure there aren't any duplicate names in the sets of built-in unit names and quantity names as well.

For your purposes, you can get a `Unit` object for AMU like so:

```
>>> from unyt import Unit
>>> amu = Unit('amu')
```
or

```
>>> from unyt import amu
[...]

PR #459, opened 2023-10-30
Handle duplicate symbols in the unyt top-level namespace

jzuhone:
This PR handles duplicate symbols from `unit_symbols` (`Unit` objects) and `physical_constants` (`unyt_quantity` objects) in the `unyt` top-level namespace.

* It removes `amu` as a physical constant, since it is typically used as a unit of measure.
* For all other duplicates in the top-level namespace, physical constants are what will be exposed. 
* Documents the fact that duplicates in the top-level namespace defer to physical constants.

Resolves Issue https://github.com/yt-project/unyt/issues/440.

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -22,8 +22,12 @@ constants defined in :mod:`unyt`, see :ref:`unit-listing`.
     :mod:`unyt` namespace. Some names occur as both unit symbols and physical
     constants, e.g. ``"me"``, ``"mp"``, ``"E_pl"``, etc. In such cases, the
     top-level namespace defaults to exporting the physical constant with the
-    duplicate name. As described below, there are other ways to import the
-    unit symbols and/or physical constants besides the top-level namespace.
+    duplicate name. This means that there may be a (very rare) case where an
+    :class:`Unit <unyt.unit_object.Unit>` object that at one point can be
+    imported from the top-level namespace may at a later point be only imported
+    as an :class:`unyt_quantity <unyt.array.unyt_quantity>` object with the
+    same name. As described below, there are other ways to import the unit
+    symbols and/or physical constants besides the top-level namespace.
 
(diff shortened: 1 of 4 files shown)


neutrinoceros (commented):
This is good. While we're on this, I guess we should define a simple but clear stability policy for unit symbols in the top level namespace: as we stand, adding a new constant always comes with the possibility of shadowing an existing unit. So, what should we do then ? I see two possible policies
- such constants won't be added to the global namespace (worse api consistency, better api stability)
- declare all unit symbols in the top level namespace unstable (better consistency, worse stability)

I think I have a slight preference for the first option, but clearly none of these is ideal. My main concern is that whichever policy is picked be explicitly stated in the docs.

jzuhone:
@neutrinoceros so I'll be honest, my preference is for the second option. It's mainly because I suspect that most constants and units that are added from now on are not likely to be duplicates, as most of the obvious ones that suffer from this have already been added. I guess in theory that means that we won't run into either option you stated, but in case it does I would err on the side of favoring constants. There is already a statement in the docs from this PR to that effect, let me know if you want more than what's said there.

neutrinoceros:
> my preference is for the second option.

Seems reasonable to me.

> There is already a statement in the docs from this PR to that effect, let me know if you want more than what's said there.

Yes, I'd like it be explicitly mentioned what's actually stable and what's not. The current phrasing already implies it, but I'd rather not leave room to interpretation.

neutrinoceros (approved):

neutrinoceros:
LGTM. Feel free to self-merge when tests complete

neutrinoceros:
apparently this broke yt's nightly tests.
https://github.com/yt-project/yt/actions/runs/6715141128/job/18249421573

Probably nothing too bad, I'm investigating it now.

Issue #462, opened 2023-11-01
BUG: (NEP 18) spurious errors when mixing array-like with unyt_array with non-default registry


neutrinoceros:
* unyt version: 3.0.0
* Python version: 3.9.17
* Operating System: any

### Description

array functions (NEP 18) are supposed to treat pure numpy arrays as dimensionless, but the way this treatment is implemented in unyt 3.0.0 isn't robust enough and it chokes when passed Python lists, or unyt_arrays that use a registry other than the default one.

### What I Did

The following 3 examples should be produce similar results and certainly not raise errors
```python
import numpy as np
from unyt.unit_object import Unit
from unyt.unit_registry import UnitRegistry

# this is ok
np.concatenate([np.array([1]), [2]*Unit()])

# but this isn't
np.concatenate([[1], [2]*Unit()])

# and neither is this
np.concatenate([np.array([1]), [2]*Unit(registry=UnitRegistry())])
```

neutrinoceros:
There are really two bugs in here, and I'll try to address them separately

PR #463, opened 2023-11-01
BUG: fix an issue where array functions would raise UnitConsistencyError on unyt arrays using non-default unit registries

neutrinoceros:
Address the first part of #462

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -216,8 +216,8 @@ def _validate_units_consistency(objs):
     # because it's already a necessary condition for numpy to use our
     # custom implementations
     units = get_units(objs)
-    sunits = set(units)
-    if len(sunits) == 1:
+    unique_units = set(units)
+    if len(unique_units) == 1 or all(u.is_dimensionless for u in unique_units):
         return units[0]
     else:
         raise UnitInconsistencyError(*units)
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -18,6 +18,8 @@ from unyt.exceptions import (
     UnytError,
 )
 from unyt.testing import assert_array_equal_units
+from unyt.unit_object import Unit
+from unyt.unit_registry import UnitRegistry
 
 NUMPY_VERSION = Version(version("numpy"))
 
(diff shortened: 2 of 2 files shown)


unyt/_array_functions.py
```diff
@@ -216,8 +216,8 @@ def _validate_units_consistency(objs):
     # because it's already a necessary condition for numpy to use our
     # custom implementations
     units = get_units(objs)
-    sunits = set(units)
-    if len(sunits) == 1:
+    unique_units = set(units)
+    if len(unique_units) == 1 or all(u.is_dimensionless for u in unique_units):
```
jzuhone, line 220: Could you explain a little bit more how this resolves the issue (sorry, it may be because it's late for me)?

jzuhone:
What happens if the unit is dimensionless but is not the `dimensionless` unit? Say, what happens if it is `Zsun`?

unyt/_array_functions.py
neutrinoceros: the problem was that 
- `Unit()` and `Unit(registry=UnitRegistry())` have different hash, so they are considered as 2 separate unique units
- `Unit()` (or `NULL_UNIT` is this context) gets automatically inserted for non-unyt array-like data structures only because we want them to be treated as dimensionless

This condition makes it so all dimensionless units are treated as equal as long as nothing else is received.

neutrinoceros:
> What happens if the unit is dimensionless but is not the dimensionless unit? Say, what happens if it is Zsun?

that's good question. I think we should imitate the behaviour of basic operations (like addition). At the moment, `1*Zsun+1` or `1*Zsun+1*dimensionless` are all treated as valid operations (in unyt 2.9.5 as well as 3.0.0). I must say I don't understand the result though, or why `1*Zsun + 1 != 1 + 1*Zsun`, is this another bug ?

jzuhone:
@neutrinoceros they are equal--I just checked:

```python
import unyt as u
a = 1+1.0*u.Zsun
b = 1.0*u.Zsun+1
print(a)
print(b)
print(a == b)
```

gives:
```pycon
1.01295 dimensionless
78.22007722007721 Zsun
True
```

PR #464, opened 2023-11-02
BUG: fix an issue where array functions would crash (AttributeError) when passed non-ndarray array like objects (e.g. Python lists)

### LINKED ISSUE #462: BUG: (NEP 18) spurious errors when mixing array-like with unyt_array with non-default registry

neutrinoceros:
based on #463, closes #462 

this patch was semi-automated with the following script:
```python
import re
from pathlib import Path

module_file = Path(__file__).parent / "unyt" / "_array_functions.py"

module_file.write_text(
    re.sub(
        r"(?P<varname>[\w_]+)\.view\(np.ndarray\)",
        lambda match: f"np.asarray({match.group('varname')})",
        module_file.read_text(),
    )
)
```

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -68,10 +68,8 @@ def array2string(a, *args, **kwargs):
 def product_helper(a, b, out, func):
     prod_units = getattr(a, "units", NULL_UNIT) * getattr(b, "units", NULL_UNIT)
     if out is None:
-        return func._implementation(a.view(np.ndarray), b.view(np.ndarray)) * prod_units
-    res = func._implementation(
-        a.view(np.ndarray), b.view(np.ndarray), out=out.view(np.ndarray)
-    )
+        return func._implementation(np.asarray(a), np.asarray(b)) * prod_units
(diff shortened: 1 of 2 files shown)


neutrinoceros:
rebased, and ready for review @jzuhone

PR #476, opened 2023-11-11
BUG: fix caching error with UnitRegistry.modify

### LINKED ISSUE #473: unable to use custom unit_registry nor modify base_value for unit in default_unit_registry
* unyt version: 2.9.5
* Python version: 3.10.1
* Operating System: macOS Ventura 13.2

### Description

I have two issues. The first is when creating a custom unit_registry I can't seem to be able to be able to push the changes in any meaningful way to use the units in said registry. The second is when adding custom units to the default unit registry, I can't modify the base_value after the fact.

### What I Did

```
import unyt

# ISSUE 1
my_registry = unyt.UnitRegistry()
unyt.unit_object.define_unit('bushel', 0.5*unyt.kg, registry=my_registry)
a = 2. * unyt.bushel

  File "/Users/adamcohan/Desktop/Clinic/pygreet/testing_units/testing_registries.py", line 5, in <module>
    a = 2. * unyt.bushel
AttributeError: module 'unyt' has no attribute 'bushel'

# ISSUE 2
unyt.unit_object.define_unit('bushel', 0.5*unyt.kg)
print(unyt.bushel.base_value)
print(unyt.unit_registry.default_unit_registry['bushel'])
unyt.unit_registry.default_unit_registry.modify('bushel', 0.1*unyt.kg)
print(unyt.bushel.base_value)
print(unyt.unit_registry.default_unit_registry['bushel'])

0.5
(0.5, (mass), 0.0, '\\rm{bushel}', False)
0.5
(0.1, (mass), 0.0, '\\rm{bushel}', False)
```

neutrinoceros:
close #473

### DIFF
--- a/unyt/tests/test_unit_registry.py
+++ b/unyt/tests/test_unit_registry.py
@@ -76,6 +76,14 @@ def test_modify_cache_clear():
     assert 1.0 * u2 == 0.5 * u0
 
 
+def test_remove_unit():
+    ureg = UnitRegistry()
+    ureg.add("celery", 1.0, length)
+    ureg.remove("celery")
+    with pytest.raises(UnitParseError):
+        Unit("celery", registry=ureg)
+
+
 def test_keys():
     ureg = UnitRegistry()
     assert sorted(ureg.keys()) == sorted(ureg.lut.keys())
--- a/unyt/tests/test_unit_registry.py
+++ b/unyt/tests/test_unit_registry.py
@@ -57,6 +57,25 @@ def test_modify_symbol_from_default_unit_registry():
         default_unit_registry.remove("cm")
 
 
+def test_modify_cache_clear():
+    # see https://github.com/yt-project/unyt/issues/473
+
+    ureg = UnitRegistry()
+    ureg.add("celery", 1.0, length)
+    u0 = Unit("m", registry=ureg)
+    u1 = Unit("celery", registry=ureg)
+    assert u1 == u0
+
+    ureg.modify("celery", 0.5)
+
+    # check that this equality still holds post registry modification
+    assert u1 == u0
+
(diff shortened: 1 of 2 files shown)


unyt/tests/test_unit_registry.py
```diff
@@ -57,6 +57,29 @@ def test_modify_symbol_from_default_unit_registry():
...
+    ureg = UnitRegistry()
+    ureg.add("celery", 1.0, length)
+    u0 = Unit("m", registry=ureg)
+    Unit("celery", registry=ureg)
+
+    ureg.modify("celery", 0.5)
+    u2 = Unit("celery", registry=ureg)
+    assert 1.0 * u2 == 0.5 * u0
+
+    # TODO: test behaviour in u1 pre/post modification
```
neutrinoceros, line 72: this line refers to the Unit created on after u0 (I guess linters removed the variable binding to `u1` because I didn't use it). The behaviour I'm getting with the branch at the moment is that the Unit object created before `ureg.modify` still uses the original base value even after modification of the registry. This seems to be in line with @AdamCohan's expectations, and seems consistent with the idea that `Unit` objects should be immutable, but I don't think that this case is currently *tested*. @jzuhone, what should be the expected behaviour in your opinion ?

neutrinoceros: Let's try that again: @jzuhone

jzuhone: The behavior you are seeing is correct.

neutrinoceros: thank you ! I've updated the test to reflect this and rebased the branch, this is now ready for review !

jzuhone (approved):

PR #478, opened 2023-11-29
BUG: fix an issue where array functions would raise UnitInconsistencyError operands' units differ by some dimensionless factor

neutrinoceros:
Adding the test first, will push the fix after CI goes red.


neutrinoceros:
@jzuhone if you agree to this fix, I think that it could justify a 3.0.2 release

jzuhone (approved):

PR #486, opened 2024-02-05
BUG: fix compat with numpy dev for np.trapezoid (previously named np.trapz)

neutrinoceros:
xref https://github.com/numpy/numpy/pull/25738


neutrinoceros:
This isn't sufficient to fix bleeding-edge CI anymore.  #489 is also needed (but independent)

jzuhone (approved):

PR #491, opened 2024-03-05
Add suport for np.cbrt

yuyttenhove:
Hi,

I noticed that currently `np.cbrt` is not supported on unyt arrays. (Getting `KeyError: <ufunc 'cbrt'>` at `unyt/array.py:1814, in unyt_array.__array_ufunc__(self, ufunc, method, *inputs, **kwargs)`)

This very small pull request adds that. I basically copied what was done for np.sqrt.

I extended the relevant unit test in `unyt/tests/test_unyt_array.py` and I don't think this part of `unyt` is documented, so I didn't add any documentation.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -26,6 +26,7 @@ from numpy import (
     bitwise_and,
     bitwise_or,
     bitwise_xor,
+    cbrt,
     ceil,
     clip,
     conj,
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Thanks @yuyttenhove, your patch looks perfect !

Before I merge, do you know of any other functions that we're missing as well ? Or did you just happen to try this one and found that it didn't work ?

neutrinoceros (approved):

yuyttenhove:
I just happened to stumble on this one and didn't notice any other missing functions during my use of unyt, but I didn't look into it extensively.

neutrinoceros:
Ok no worries. If you ever find anything else, these patches are always welcome. Thanks again !

PR #498, opened 2024-03-29
TST: drop test case for unpickling old pickle files as too sensitive to upstream changes

### LINKED ISSUE #497: TST: upcoming test failure in bleeding-edge CI
bleeding-edge CI is red again, [example log](https://github.com/yt-project/unyt/actions/runs/8460322825)
This time, the "culprit" seems to be sympy 1.12.1a1

However, I'm hesitant to report this upstream: the one failing test was originally added because of a similar "regression" where newest sympy broke unpickling old pickle files. The reason why we have this test is that `yt`'s CI relies on pickle files to store known-good test results. However, pickles were never intended to be a reliable long term storage solution, so I don't think it's fair to put the burden of maintaining this system onto sympy maintainers, and unyt is only a middle-man with no actual responsibility or power over this, so I think our best shot is to actually kill the test.

That said, I will still warn other yt maintainers about the upcoming expiration of some test answers.

neutrinoceros:
closes #497


neutrinoceros:
Ping @jzuhone for review

PR #506, opened 2024-06-17
DOC: adjust doctests to changes in array repr from numpy 2.0

neutrinoceros:
A couple non-user facing adjustments to numpy 2.0

[example logs](https://github.com/yt-project/unyt/actions/runs/9540016330/job/26291141524)

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -1494,7 +1494,7 @@ Methods that hang off of a :mod:`unyt_dask_array` object and operations on
 One important caveat is that using Dask array functions may strip units:
 
     >>> da.sum(x_da).compute()
-    49995000
+    np.int64(49995000)
 
 For simple reductions, you can use the :mod:`reduce_with_units` function:
 
--- a/tox.ini
+++ b/tox.ini
@@ -21,7 +21,7 @@ deps =
     pytest
     h5py
     pint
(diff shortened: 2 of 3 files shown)


neutrinoceros:
I don't get why CI is red yet. Even on the failing job it looks like all tests passed. Let me give it another try before I dive into this potential rabbit hole.

neutrinoceros:
I see GitHub Actions re-released several Python binaries in the last couple days, so let's try again just again the problem is already resolved.

neutrinoceros:
Alright, new problem: now Windows is failing because it doesn't yet get to install numpy 2. This is caused by the latest version of astropy (6.1.1) pinning `numpy<2.0` specifically on windows (and I'm really glad I *authored* this pin, otherwise I'm not sure I would have found out)
I also know this will be resolved in astropy 6.1.2 (coming soon), and that 6.1.0 didn't have the pin at all. Since I don't think astropy bugs discovered/patched since 6.1.0 actually affect unyt I'm just going to try and exclude 6.1.1 and see how it goes.

neutrinoceros:
turns out I was just not scanning logs correctly and there was an actual failure on Python 3.9 with minimal supported numpy. Trying to fix this now.

PR #512, opened 2024-07-03
BUG: fix incompatibilities with NumPy 2.1

neutrinoceros:
- **BUG: test/implement array functions new in NumPy 2.1 (np.cumulative_sum and np.cumulative_prod)**
- **BUG: adapt np.clip's implementation signature at runtime**

[example logs](https://github.com/yt-project/unyt/actions/runs/9774860392/job/26984040518)
xref: https://github.com/numpy/numpy/pull/26724

### DIFF
--- a/docs/usage.rst
+++ b/docs/usage.rst
@@ -269,10 +269,9 @@ controlled identically to NumPy arrays, using ``numpy.setprintoptions``:
   >>> import numpy as np
   >>> import unyt as u
   ...
-  >>> np.set_printoptions(precision=4)
-  >>> print([1.123456789]*u.km)
+  >>> with np.printoptions(precision=4):
+  ...     print([1.123456789]*u.km)
   [1.1235] km
-  >>> np.set_printoptions(precision=8)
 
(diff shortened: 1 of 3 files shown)


neutrinoceros:
Normal CI is affected now that NumPy 2.1 is out (though it is not the only issue)
@jzuhone ?

docs/usage.rst
```diff
@@ -269,10 +269,9 @@ controlled identically to NumPy arrays, using ``numpy.setprintoptions``:
   >>> import numpy as np
   >>> import unyt as u
   ...
-  >>> np.set_printoptions(precision=4)
-  >>> print([1.123456789]*u.km)
+  >>> with np.printoptions(precision=4):
+  ...     print([1.123456789]*u.km)
```
neutrinoceros, line 273: the reason this fix is needed is that the repr of `np.set_printoptions`'s return value is not empty in numpy 2.1, which caused this test to fail *and* leak side effects, causing a lot *other* tests to fail too.

jzuhone (approved):

PR #515, opened 2024-08-19
TST: enable doctests in bleeding-edge CI

neutrinoceros:
NumPy 2.1 was just released and appears to break a bunch of doctests and I didn't this that coming. With this patch, this could be caught in advance next time. I'll try fixing the issue too but maybe that'll be a different PR depending on how heavy the fix.
Also note that problems solved in #512 will also fail CI here.


neutrinoceros:
Actually I remember now why these tests are not included here: we have no easy way to skip tests that require otherwise optional dependencies, and I don't want to wait on those to check compat with core deps.

PR #520, opened 2024-09-28
Fix Stephan-Boltzmann constant accuracy and add radiation constant

zingale:
and make it and Stefan-Boltzmann derived from fundamental units


unyt/equivalencies.py
```diff
@@ -473,9 +473,9 @@ class EffectiveTemperatureEquivalence(Equivalence):
     effective_temperature: flux <-> temperature
     >>> from unyt import K, W, m
     >>> (5000.*K).to_equivalent('W/m**2', 'effective_temperature')
-    unyt_quantity(35439831.25, 'W/m**2')
+    unyt_quantity(35439828.89119583, 'W/m**2')
```
neutrinoceros, line 476: I'm not sure such a large change is warranted (unless it somehow counts as a bug fix), but why is this output changing at all ?

neutrinoceros (commented):
I'm not against it, though I may have some reserves about using a single greek letter symbol for such a domain specific constant, but I do not understand why it's having an effect on existing doctests. Can you clarify ?

zingale:
I didn't change the name of the constant -- it was already σ in the code.

What I did is fix the way the constant is computed.  The Stefan-Boltzmann constant is not a fundamental constant -- it is constructed from Planck's constant, Boltzmann's constant and the speed of light (https://en.wikipedia.org/wiki/Stefan%E2%80%93Boltzmann_law#Stefan%E2%80%93Boltzmann_constant).  So now its value is consistent with the values of the fundamental constants that are already in unyt.

The reason for the doctest change is that σ's value changed slightly, because it is now defined in terms of the fundamental constants.

neutrinoceros:
> I didn't change the name of the constant -- it was already σ in the code.

Whoops, sorry I misread the diff

Ok thanks for clarifying, I missed that there were actually two changes here: fixing `stefan_boltzmann_constant` and adding `radiation_constant`.
Searching for the latter, I find https://en.wikipedia.org/wiki/Radiation_constant, indicating that the name is somewhat ambiguous. Could we use `stephan_boltzmann_constant` or `radiation_density_constant` instead ?

zingale:
radiation density constant is fine, although I've never seen it referred to anything other than the "radiation constant" in textbooks.

neutrinoceros:
> radiation density constant is fine

Alright, let's go with that then.

zingale:
I've updated the name to `radiation_density_constant`

neutrinoceros:
pre-commit.ci autofix

neutrinoceros:
Thanks @zingale ! FYI I'm planning to release this into unyt 3.0.4 which I'll hopefully be able to prepare sometimes next week.

PR #521, opened 2024-09-28
add foe and Bethe units

zingale:
these are used in supernovae

see: https://en.wikipedia.org/wiki/Foe_(unit)

### DIFF
--- a/unyt/_unit_lookup_table.py
+++ b/unyt/_unit_lookup_table.py
@@ -429,6 +429,7 @@ default_unit_symbol_lut = OrderedDict(
         ),
         # misc
         ("eV", (J_per_eV, dimensions.energy, 0.0, r"\rm{eV}", True)),
+        # note: 1 foe = 1 Bethe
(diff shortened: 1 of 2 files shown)


unyt/_unit_lookup_table.py
```diff
@@ -428,6 +429,8 @@
         ),
         # misc
         ("eV", (J_per_eV, dimensions.energy, 0.0, r"\rm{eV}", True)),
+        ("foe", (J_per_foe, dimensions.energy, 0.0, r"\rm{foe}", False)),
+        ("bethe", (J_per_foe, dimensions.energy, 0.0, r"\rm{B}", False)),
```
neutrinoceros, line 433: This is correct, but *looks* like a mistake. Can you add a comment to reassure future readers that yes, 1 foe = 1 bethe ?

zingale: comment added

PR #524, opened 2024-10-08
BUG: fix raising a unyt array to an array power in sensible cases

### LINKED ISSUE #522: An error occurs when the exponent is an array
* unyt version: 2.9.5
* Python version: 3.9.17
* Operating System: 14.6.1 

### Description

In the `power` operation of unyt arrays, when the exponent is also an array, it leads to an error.

```shell
In [1]: from unyt import unyt_array

In [2]: a = unyt_array([1,2,3,4])

In [3]: b = unyt_array([1,2,3,4])

In [4]: a ** b
ValueError                                Traceback (most recent call last)
Cell In[4], line 1
----> 1 a ** b

File ~/Apps/anaconda3/lib/python3.9/site-packages/unyt/array.py:1745, in unyt_array.__pow__(self, p, mod)
   1737 def __pow__(self, p, mod=None, /):
   1738     """
   1739     Power function, over-rides the ufunc as
   1740     ``numpy`` passes the ``_ones_like`` ufunc for
   1741     powers of zero (in some cases), which leads to an
   1742     incorrect unit calculation.
   1743     """
-> 1745     if p == 0.0:
   1746         ret = self.ua
   1747         ret.units = Unit("dimensionless")

ValueError: The truth value of an array with more than one element is ambiguous. Use a.any() or a.all()
```

While for numpy arrays, it behaviors like 

```shell
In [5]: import numpy as np

In [6]: a = np.array([1,2,3,4])

In [7]: b = np.array([1,2,3,4])

In [8]: a ** b
Out[8]: array([  1,   4,  27, 256])
```

I now use `to_ndarray()` to make it work but lose the units.

neutrinoceros:
Trying to fix the few cases where it *does* make sense to raise a unyt array to a power array.
Close #522

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1764,8 +1764,8 @@ class unyt_array(np.ndarray):
         Power function
         """
         # see https://github.com/yt-project/unyt/issues/203
-        if p == 0.0:
-            ret = self.ua
+        if np.isscalar(p) and p == 0.0:
+            ret = self.unit_array
             ret.units = Unit("dimensionless")
             return ret
         else:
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Yes, if the power itself is dimensionless, it should work. See #524 and please let me know if this helps !

xshaokun (commented):
I’m sorry for the delayed response. When 
```python
a = unyt_array([1, 2, 3])
p = unyt_array([1, 2, 3, 4])
```
`a**p` should not raise an error. However, I tried it and that’s not the case.

chrishavlin:
> a**p should not raise an error. However, I tried it and that’s not the case.

What's your expectation when the array shapes don't match? In your example, those arrays will raise a broadcast error with plain numpy, so I wouldn't expect the operation to work with unyt. 

```
a = np.array([1, 2, 3])
p = np.array([1, 2, 3, 4]
a**p
```
```python-traceback
ValueError: operands could not be broadcast together with shapes (3,) (4,) 
```

xshaokun:
> > a**p should not raise an error. However, I tried it and that’s not the case.
> 
> 
> 
> What's your expectation when the array shapes don't match? In your example, those arrays will raise a broadcast error with plain numpy, so I wouldn't expect the operation to work with unyt. 
> 
> 
> 
> ```
> 
> a = np.array([1, 2, 3])
> 
> p = np.array([1, 2, 3, 4]
> 
> a**p
> 
> ```
> 
> ```python-traceback
> 
> ValueError: operands could not be broadcast together with shapes (3,) (4,) 
> 
> ```

Oh, I'm sorry for the typo! These two arrays are supposed to be the same, or any two dimensionless unyt_arrays with the same shape.

xshaokun:
Thank you so much for your work, everything is working fine now. Since I don’t fully understand some of your implements, I’m unable to suggest ways to simplify the logic. Nevertheless, I really appreciate it!

PR #525, opened 2024-10-08
TST: add CPython 3.13 to regular test matrix

neutrinoceros:
blocked by:
- https://github.com/hgrecco/pint/issues/1969

### DIFF
--- a/.github/workflows/ci.yml
+++ b/.github/workflows/ci.yml
@@ -22,6 +22,7 @@ jobs:
           - '3.10'
           - '3.11'
           - '3.12'
+          - '3.13'
         # Test all on ubuntu, test ends on macos and windows
         include:
           - os: macos-13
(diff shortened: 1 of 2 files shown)


neutrinoceros:
Looks like we're hitting a known issue from pint
xref https://github.com/hgrecco/pint/pull/2037

I don't want to be blocked by development of an optional dependency that we merely want to interface with, so I'll be looking for a way to skip pint tests on cp313 for now.

neutrinoceros:
I actually don't know of a good solution. The only one I can think of is skipping *all* doctests on py313 for now, but then it's not ready to replace py312 entierely on CI and I don't want to add jobs just to get around a bug in pint. So, unless the upstream issue is promptly resolved, I'll just not include this in uny 3.0.4

neutrinoceros:
pint 0.24.4 was just released, let's try this again

PR #528, opened 2024-10-22
MNT: switch Python 3.9 + mac testing from macos-12 to macos-13

neutrinoceros:
Following the upcoming deprecation of this runner image https://github.com/actions/runner-images/issues/10721


neutrinoceros:
we're hitting the brownout period, meaning jobs are starting to randomly fail because of this, so I'll exceptionally self-merge here.

PR #532, opened 2024-10-29
DEP: make pyproject.toml the one source of truth for test-only dependencies"

neutrinoceros:
This PR does 3 things:

### Dependency groups

Move dev-only dependency specifications from `tox.ini` to `pyproject.toml` as [PEP 735 Dependency Groups](https://peps.python.org/pep-0735/).

The PEP was accepted a couple weeks ago and is already supported by tox *as well as uv*, which allow running tests without tox as, for instance

```shell
uv run --group test pytest
```

### Rely on `uv`'s dependency resolver to test with pinned versions

Avoid duplication of metadata for oldest supported versions of our dependencies by switching dependency resolution strategy. This is supported by `uv` (via `tox-uv`).
Notes:
- this also means that test dependencies (e.g. pytest) are pinned as well (good)
- using this strategy requires setting explicit minimal requirements everywhere (otherwise one gets the oldest version in existence, which in a very general way is completely broken)
- in order to properly test this on my setup (Apple Silicon), I needed to add a layer of requirements specifically for this platform (as our absolute-oldest supported versions work on ubuntu but are generally lacking support for Apple Silicon)

## Drop pytest-cov

I removed `pytest-cov` from our testing pipeline, as it's a completely unnecessary layer of indirection between `pytest` and `coverage`. I switched to direct invokes to `coverage`.

While I was at it, I also re-organized `tox.ini` a bit (mostly moving lines around).


neutrinoceros:
Apparently something I did broke tox env selection in CI so every job runs every env. I don't know yet if that's the reason it's failing, or if that's a different problem.

neutrinoceros:
I resolved the problem mentioned above by using `tox-gh` (first party plugin) instead of `tox-gh-actions` (third party plugin). This could probably be avoided or simplified but I already spent > 3hrs on this so I'll quit while I'm ahead.

neutrinoceros:
an interesting side effect of using tox-uv in CI is that it saves a surprising amount of CPU time.
See for instance usage pages
- [before](https://github.com/yt-project/unyt/actions/runs/11564973198/usage) (total CPU time: 16m23s)
- [after](https://github.com/yt-project/unyt/actions/runs/11573470309/usage?pr=532) (total CPU time: 11m2s)

jzuhone:
@neutrinoceros can you fix the conflicts here?

neutrinoceros:
done.

neutrinoceros:
(Looks like I'm going to need a couple pushes before this is stable again, so, drafting in the mean time)

neutrinoceros:
hum, I think some of our *indirect* dependencies are now preventing use of our prefered numpy in `py39-versions`.
Locking dependencies with `uv lock --resolution lowest-direct` produces
```toml
[[package]]
...
dependencies = [
    ...,
    { name = "numpy", version = "1.21.3", source = { registry = "https://pypi.org/simple" }, marker = "python_full_version < '3.10' and platform_machine == 'arm64' and sys_platform == 'darwin'" },
    { name = "numpy", version = "1.23.0", source = { registry = "https://pypi.org/simple" }, marker = "(python_full_version < '3.10' and platform_machine != 'arm64') or (python_full_version < '3.10' and sys_platform != 'darwin')" },
    ...,
]
```

while `pyproject.toml` has
```toml
[project]
name = "bokeh"
...
dependencies = [
    ...,
    "numpy>=1.19.3, <3.0",
    "numpy>=1.21.3, <3.0 ; platform_machine=='arm64' and platform_system=='Darwin'",
    ...,
]
```

The solution would be to add resolution constraints, but it might be much easier to tackle once we drop support for Python 3.9, which will naturally resolve a bunch of forks. I propose to post-pone this until after unyt 3.0.4 then.

neutrinoceros:
This isn't blocked by any other PR anymore but it's now in a broken state, and it's not immediately clear what's the broken component so I think I'll salvage my work by splitting this PR.

neutrinoceros:
ok I moved the most immediately useful bits to the following PRs
- #574
- #575

what's still exclusive to this PR (and potentially broken):
- using `uv` in tox, which allows de-duplication of lower-bounds in all optional dependencies with `--resolution=lowest-direct`)
- migrating from `tox-gh-actions` to `tox-gh` (not sure why I did that originally)

PR #537, opened 2024-12-15
Prod bugfix

### LINKED ISSUE #536: Incorrect units when using `np.prod` with an `axis` kwarg
* unyt version: 3.0.3.post34+gecd71ea
* Python version: 3.12
* Operating System: Linux (Fedora 40)

### Description

There is a bug in the unyt implementation of the numpy `prod` function. It does not correctly set the units in the returned array when an axis keyword is passed.

### What I Did

```
>>> import unyt as u
>>> np.prod(np.arange(1,10).reshape((3,3)) * u.cm,axis=0)
unyt_array([ 28,  80, 162], 'cm**9')
```
Expected units are `cm**3` in this case.

kyleaoman:
Closes #536 

When an `axis` keyword is passed to `np.prod`, the power that the dimensions should be raised to are not the size of the array but the ratio of the sizes of the input array and returned array (it's always this, but without the kwarg the size of the returned array is 1).

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -535,7 +535,8 @@ def copyto(dst, src, *args, **kwargs):
 
 @implements(np.prod)
 def prod(a, *args, **kwargs):
-    return np.prod._implementation(np.asarray(a), *args, **kwargs) * a.units**a.size
+    res = np.prod._implementation(np.asarray(a), *args, **kwargs)
+    return res * a.units**(a.size // res.size)
 
 
 @implements(np.var)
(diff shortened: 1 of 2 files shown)


kyleaoman:
pre-commit.ci autofix

neutrinoceros (commented):
Thanks for noticing *and* patching this !
The same problem is present with `np.nanprod`, and your patch *also* fixes it, however that's mostly due to an implementation detail in `numpy` that we don't control so it should be tested too.
Otherwise, LGTM

kyleaoman:
@neutrinoceros done - just parametrized the new test across `prod` and `nanprod`.

neutrinoceros (approved):

neutrinoceros:
Thanks ! let's get this one in !

PR #539, opened 2024-12-15
BUG: `np.histogram*` functions give correct units when `weights` and/or `density` are set

### LINKED ISSUE #538: Incorrect units or crash in histograms when density and/or weights present
* unyt version: 3.0.3.post34+gecd71ea
* Python version: 3.12
* Operating System: Linux (Fedora 40)

### Description

When making a histogram with `density=True` I expect the histogram to have units equal to the inverse of those of the input array.

When making a histogram with `weights=...` and the weights have units, I expect the histogram to have units equal to the units of the weights (if they have any).

When both `weights=...` and `density=True` are given I expect the combination of the above.

These principles should also extend to 2d and dd histograms.

### What I Did

The basic `histogram` just doesn't apply any units to the first return value:
```
>>> np.histogram(np.arange(10) * u.cm, density=True, weights=np.arange(10) * u.s)
(array([0.        , 0.02469136, 0.04938272, 0.07407407, 0.09876543,
       0.12345679, 0.14814815, 0.17283951, 0.19753086, 0.22222222]), unyt_array([0. , 0.9, 1.8, 2.7, 3.6, 4.5, 5.4, 6.3, 7.2, 8.1, 9. ], 'cm'))
```
The `histogram2d` actually crashes when weights with units are given:
```
>>> np.histogram2d(np.arange(10) * u.cm, np.arange(10) * u.g, density=True, weights=np.arange(10) * u.s)
  File "/home/txwx36/code/unyt/unyt/array.py", line 2045, in __array_function__
    return _HANDLED_FUNCTIONS[func](*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/txwx36/code/unyt/unyt/_array_functions.py", line 179, in histogram2d
    counts, xbins, ybins = np.histogram2d._implementation(
                           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/txwx36/.venv/koman/lib64/python3.12/site-packages/numpy/lib/_twodim_base_impl.py", line 813, in histogram2d
    hist, edges = histogramdd([x, y], bins, range, density, weights)
                  ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/txwx36/code/unyt/unyt/array.py", line 2045, in __array_function__
    return _HANDLED_FUNCTIONS[func](*args, **kwargs)
           ^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^^
  File "/home/txwx36/code/unyt/unyt/_array_functions.py", line 187, in histogramdd
    units = [_.units for _ in sample]
             ^^^^^^^
AttributeError: 'numpy.ndarray' object has no attribute 'units'
```

kyleaoman:
Closes #538 

When `weights` are given to a histogram function the histogram should inherit the units of the weights. When `density=True` is set, the histogram should have the inverse units of each dimension of the input applied.

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -202,7 +202,9 @@ def histogram2d(x, y, bins=10, range=None, density=None, weights=None, *args, **
 
 
 @implements(np.histogramdd)
-def histogramdd(sample, bins=10, range=None, density=None, weights=None, *args, **kwargs):
+def histogramdd(
+    sample, bins=10, range=None, density=None, weights=None, *args, **kwargs
+):
     units = [_.units for _ in sample]
     range = _sanitize_range(range, units=units)
     counts, bins = np.histogramdd._implementation(
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -158,20 +158,25 @@ def _sanitize_range(_range, units):
     return new_range.squeeze()
 
 
-@implements(np.histogram)
-def histogram(
-    a,
-    bins=10,
-    range=None,
-    density=None,
-    weights=None,
-    *args,
-    **kwargs,
-):
+def _histogram(a, bins=10, range=None, density=None, weights=None, normed=None):
     range = _sanitize_range(range, units=[a.units])
-    counts, bins = np.histogram._implementation(
-        np.asarray(a), bins, range, *args, **kwargs
-    )
+    if NUMPY_VERSION > Version("1.23"):
+        counts, bins = np.histogram._implementation(
+            np.asarray(a),
+            bins=bins,
+            range=range,
+            density=density,
+            weights=np.asarray(weights) if weights is not None else None,
+        )
+    else:
+        counts, bins = np.histogram._implementation(
+            np.asarray(a),
+            bins=bins,
+            range=range,
+            normed=normed,
+            weights=np.asarray(weights) if weights is not None else None,
+            density=density,
+        )
     # a and/or weights could have units, only apply if present
     # don't getattr(..., "units", NULL_UNIT) because e.g. we don't want
     # a unyt_array if weights are not a unyt_array and not density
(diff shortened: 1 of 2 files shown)


kyleaoman:
pre-commit.ci autofix

unyt/_array_functions.py
```diff
@@ -163,32 +163,62 @@ def histogram(
     a,
     bins=10,
     range=None,
+    density=None,
+    weights=None,
     *args,
     **kwargs,
```
neutrinoceros, line 169: there is a backward compatibility issue with this signature (seen in CI): numpy versions `1.19` to `1.23`, which are still supported by `unyt`, also support a `normed` argument that can be either positional of keyword, and inconviently comes *before* `density`. If we want to take `density` into account, we'll need to have a different signature depending on the version of numpy that's available at runtime.
The `NUMPY_VERSION` module constant serves this purpose, see where it's already used for examples !

kyleaoman: Ah ok I was scratching my head at what was causing the failures, fixed now, hopefully in a sensible way. At least, it will be easy to clean up once numpy versions with `normed` drop out of support.

unyt/_array_functions.py
```diff
@@ -158,40 +158,163 @@ def _sanitize_range(_range, units):
...
-    range=None,
-    *args,
-    **kwargs,
-):
+def _histogram(a, bins=10, range=None, density=None, weights=None, normed=None):
     range = _sanitize_range(range, units=[a.units])
-    counts, bins = np.histogram._implementation(
-        np.asarray(a), bins, range, *args, **kwargs
-    )
+    if NUMPY_VERSION > Version("1.23"):
```
neutrinoceros, line 163:
```suggestion
    if NUMPY_VERSION >= Version("1.24"):
```

kyleaoman: And done.

neutrinoceros:
Thanks, seems to be working on all tested envs now, so that's progress ! I'll try to give a more in depth review later this week. In the meantime, could you edit the title of the PR to something more descriptive ? We include PR titles directly in release notes.

kyleaoman:
pre-commit.ci autofix

unyt/_array_functions.py
```diff
@@ -158,38 +158,160 @@ def _sanitize_range(_range, units):
...
-
-@implements(np.histogramdd)
-def histogramdd(sample, bins=10, range=None, *args, **kwargs):
-    units = [_.units for _ in sample]
-    range = _sanitize_range(range, units=units)
-    counts, bins = np.histogramdd._implementation(
-        [np.asarray(_) for _ in sample], bins, range, *args, **kwargs
-    )
-    return counts, tuple(_bin * u for _bin, u in zip(bins, units))
+def _histogram(a, bins=10, range=None, density=None, weights=None, normed=None):
```
neutrinoceros, line 161: Since this is a private function, let's make the signature robust against dev mistakes
```suggestion
def _histogram(a, *, bins=10, range=None, density=None, weights=None, normed=None):
```

unyt/tests/test_array_functions.py
```diff
@@ -663,6 +663,36 @@ def test_histogram_implicit_units():
     assert bins.units == arr.units
 
 
+def test_histogram_with_weights_density():
```
neutrinoceros, line 666: This looks like 4 separate tests to me. Could you split them ? feel free to regroup them under a test class if you like that style.

neutrinoceros (changes requested):
I think this is almost good to go as is, thanks again ! only a couple small suggestions

neutrinoceros:
pre-commit.ci autofix

neutrinoceros (approved):

PR #548, opened 2024-12-19
BUG: `apply_over_axes` no longer assumes user-supplied function preserves units

### LINKED ISSUE #545: Should `apply_over_axes` "ignore" units?
This is a question more than a bug.

The current unyt implementation of numpy's `apply_over_axes` is:

```
@implements(np.apply_over_axes)
def apply_over_axes(func, a, axes):
    res = func(np.asarray(a), axes[0]) * a.units
    if len(axes) > 1:
        # this function is recursive by nature,
        # here we intentionally do not call the base _implementation
        return np.apply_over_axes(func, res, axes[1:])
    else:
        return res
```

This assumes that the function `func` preserves the units of the input `a`. It looks like it's necessary for `func` to return values with common units, but I guess those don't strictly need to be the same as those of `a`? As a trivial example, shouldn't:
```
def f(x):
    return x ** 2
```
be allowed? In this case the result will be incorrect. Is the following implementation enough?

```
@implements(np.apply_over_axes)
def apply_over_axes(func, a, axes):
    res = func(a, axes[0])
    if len(axes) > 1:
        # this function is recursive by nature,
        # here we intentionally do not call the base _implementation
        return np.apply_over_axes(func, res, axes[1:])
    else:
        return res
```

kyleaoman:
Previously the numpy function `apply_over_axes` assumed that the first argument, a user-supplied function, returned values with the same units as the input, which could give incorrect output (e.g. for a function `lambda x, axis: x ** 2`). It's beyond the scope and ability of `unyt` to strictly enforce consistency here, but the updated function now allows units to propagate naturally through the supplied function.

Closes #545

### DIFF
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -1657,9 +1657,9 @@ def test_apply_along_axis():
 @pytest.mark.parametrize(
     "axes_expectunits",
     (
-        ((0, 1), cm ** 4),
-        ((0, ), cm ** 2),
-    )
+        ((0, 1), cm**4),
+        ((0,), cm**2),
+    ),
 )
 def test_apply_over_axes(axes_expectunits):
     axes, expectunits = axes_expectunits
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -1654,15 +1654,23 @@ def test_apply_along_axis():
     assert ret.units == cm**2
 
 
-def test_apply_over_axes():
+@pytest.mark.parametrize(
+    "axes_expectunits",
+    (
+        ((0, 1), cm ** 4),
+        ((0, ), cm ** 2),
+    )
+)
+def test_apply_over_axes(axes_expectunits):
+    axes, expectunits = axes_expectunits
     # the user-supplied function must be trusted to treat units
     # sensibly (mainly that it doesn't give a mix of units across
     # the resulting array), but we can check that units are
     # propagated correctly for well-behaved functions.
     a = np.eye(3) * cm
-    ret = np.apply_over_axes(lambda x, axis: x ** 2, a, (0, 1))
-    assert type(ret) is unyt_array
-    assert ret.units == cm**4
+    ret = np.apply_over_axes(lambda x, axis: x[axis] ** 2, a, axes)
+    assert isinstance(ret, unyt_array)  # could be subclass unyt_quantity
+    assert ret.units == expectunits
 
 
 def test_array_equal():
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -1655,10 +1655,14 @@ def test_apply_along_axis():
 
 
 def test_apply_over_axes():
+    # the user-supplied function must be trusted to treat units
+    # sensibly (mainly that it doesn't give a mix of units across
+    # the resulting array), but we can check that units are
+    # propagated correctly for well-behaved functions.
     a = np.eye(3) * cm
-    ret = np.apply_over_axes(lambda x, axis: x * cm, a, (0, 1))
+    ret = np.apply_over_axes(lambda x, axis: x ** 2, a, (0, 1))
     assert type(ret) is unyt_array
-    assert ret.units == cm**3
+    assert ret.units == cm**4
 
 
 def test_array_equal():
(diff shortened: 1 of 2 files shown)


unyt/tests/test_array_functions.py
```diff
@@ -1655,10 +1655,14 @@ def test_apply_along_axis():
 
 
 def test_apply_over_axes():
+    # the user-supplied function must be trusted to treat units
+    # sensibly (mainly that it doesn't give a mix of units across
+    # the resulting array), but we can check that units are
+    # propagated correctly for well-behaved functions.
     a = np.eye(3) * cm
-    ret = np.apply_over_axes(lambda x, axis: x * cm, a, (0, 1))
+    ret = np.apply_over_axes(lambda x, axis: x ** 2, a, (0, 1))
```
neutrinoceros, line 1663: I'm a bit annoyed to see that `axis` isn't used at all at the moment (not your fault, obviously)
```suggestion
    ret = np.apply_over_axes(lambda x, axis: x[axis] ** 2, a, (0, 1))
```

kyleaoman:
pre-commit.ci autofix

unyt/tests/test_array_functions.py
```diff
@@ -1654,11 +1654,23 @@ def test_apply_along_axis():
...
-def test_apply_over_axes():
+@pytest.mark.parametrize(
+    "axes_expectunits",
+    (
+        ((0, 1), cm**4),
+        ((0,), cm**2),
+    ),
+)
+def test_apply_over_axes(axes_expectunits):
+    axes, expectunits = axes_expectunits
```
neutrinoceros, line 1665:
```suggestion
@pytest.mark.parametrize("axes, expected_units", [((0, 1), cm**4), ((0,), cm**2)])
def test_apply_over_axes(axes, expected_units):
```

unyt/tests/test_array_functions.py
```diff
@@ -1654,11 +1654,23 @@ def test_apply_along_axis():
...
+    # sensibly (mainly that it doesn't give a mix of units across
+    # the resulting array), but we can check that units are
+    # propagated correctly for well-behaved functions.
     a = np.eye(3) * cm
-    ret = np.apply_over_axes(lambda x, axis: x * cm, a, (0, 1))
-    assert type(ret) is unyt_array
-    assert ret.units == cm**3
+    ret = np.apply_over_axes(lambda x, axis: x[axis] ** 2, a, axes)
+    assert isinstance(ret, unyt_array)  # could be subclass unyt_quantity
+    assert ret.units == expectunits
```
neutrinoceros, line 1673:
```suggestion
    assert ret.units == expected_units
```

neutrinoceros (changes requested):
almost there

kyleaoman:
Did not know about that syntax for parametrize - cool!

unyt/tests/test_array_functions.py
```diff
@@ -1654,11 +1654,16 @@ def test_apply_along_axis():
...
+    # the user-supplied function must be trusted to treat units
+    # sensibly (mainly that it doesn't give a mix of units across
+    # the resulting array), but we can check that units are
+    # propagated correctly for well-behaved functions.
     a = np.eye(3) * cm
-    ret = np.apply_over_axes(lambda x, axis: x * cm, a, (0, 1))
-    assert type(ret) is unyt_array
-    assert ret.units == cm**3
+    ret = np.apply_over_axes(lambda x, axis: x[axis] ** 2, a, axes)
+    assert isinstance(ret, unyt_array)  # could be subclass unyt_quantity
```
neutrinoceros, line 1665: Can you elaborate on this comment ? I don't think I understand what situation you're hinting at

kyleaoman: `assert type(ret) is unyt_array` breaks if `ret` is a `unyt_quantity`, which it is (correctly) in one of the test cases.

neutrinoceros: oh, I forgot that we had two classes there 🤦🏻‍♂️ 
in that case, would it make sense to rewrite the comment as follow ?
```suggestion
    assert isinstance(ret, unyt_array)  # could be unyt_quantity
```

kyleaoman: That subclass has caused me considerable consternation (until deciding to handle it properly), it is not easily forgotten ;)

neutrinoceros (approved):

PR #544, opened 2024-12-18
BUG: `linspace` and `logspace` give incorrect results or crash with some inputs

kyleaoman:
The implementations of `linspace` and `logspace` are improved so that `retstep` can be used with `linspace` and `logspace` rejects inputs for `start` and `stop` with non-dimensionless units, and also correctly handles the `base` keyword argument with units.

Tests are updated to reflect changes.

Closes #542
Closes #543

PR #553, opened 2024-12-25
BUG: fix a regression where `np.linspace`'s num argument would be ignored for `unyt_array`s

neutrinoceros:
This fixes a regression I overlooked when reviewing #544 but poped up in yt's test suite https://github.com/yt-project/yt/issues/5091

for posterity: this regression never landed in a release.


neutrinoceros:
Since the fix is so trivial (I really should have caught the bug in my review) and the bug affects downstream, I'll exceptionally self-merge for this one.

PR #557, opened 2025-01-31
RFC: handle `np.vecdot` as a ufunc rather than an arrayfunc

kyleaoman:
In the tests of array functions, one of the tested functions `np.vecdot` is actually a ufunc. I'm relying on the set of `NOOP_FUNCTIONS` in an in-development addition to the https://github.com/SWIFTSIM/swiftsimio test suite to check that our code at least supports all numpy functions supported by unyt, and having a (single) ufunc in the list breaks things. I'm therefore proposing to remove it from the list here. I've also removed the test of the function, perhaps it should be migrated to `test_unyt_array.py`, but I can't spot an obvious place to put it.

Note that `np.linalg.vecdot` *is* an array function, not a ufunc.

### DIFF
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -589,7 +589,7 @@ def test_matrix_transpose(namespace):
 @pytest.mark.skipif(
     NUMPY_VERSION < Version("2.0.0dev0"), reason="vecdot is new in numpy 2.0"
 )
-def test_vecdot(namespace):
+def test_vecdot():
     a = np.arange(0, 9)
     assert_array_equal_units(np.linalg.vecdot(a, a), np.vdot(a, a))
 
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -587,9 +587,9 @@ def test_matrix_transpose(namespace):
 
 
 @pytest.mark.skipif(
-    NUMPY_VERSION < Version("2.0.0dev0"), reason="vecdot is new in numpy 2.0"
+    NUMPY_VERSION < Version("2.0.0dev0"), reason="linalg.vecdot is new in numpy 2.0"
 )
-def test_vecdot():
+def test_linalg_vecdot():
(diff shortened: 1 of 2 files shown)


unyt/tests/test_array_functions.py
```diff
@@ -590,14 +589,9 @@ def test_matrix_transpose(namespace):
 @pytest.mark.skipif(
     NUMPY_VERSION < Version("2.0.0dev0"), reason="vecdot is new in numpy 2.0"
 )
-@pytest.mark.parametrize("namespace", [None, "linalg"])
-def test_vecdot(namespace):
-    if namespace is None:
-        func = np.vecdot
-    else:
-        func = getattr(np, namespace).vecdot
+def test_vecdot():
```
neutrinoceros, line 592: I guess this test should be renamed `test_linalg_vecdot` and moved closer to its relatives `test_linalg_*`

kyleaoman: Made the name change and updated the skipif. As to moving it, its current neighbours are `test_linalg_outer, test_linalg_cross, test_linalg_matmul, test_linalg_matrix_norm, test_matrix_transpose, test_linag_vecdot, test_linalg_vector_norm, test_linalg_svd, test_linalg_svdvals`, so there doesn't seem to be a strong case to move it.

kyleaoman:
Ah perhaps don't merge this just yet - have discovered that `vecdot` seems to be an unhandled ufunc:
```
>           unit_operator = self._ufunc_registry[ufunc]
E           KeyError: <ufunc 'vecdot'>
```

unyt/tests/test_array_functions.py
neutrinoceros: oh, I missed that it was *already* where it should. Sorry !

kyleaoman:
@neutrinoceros Ok I think that I've registered vecdot as a ufunc with appropriate guards on numpy version. Ready for a look when you have time.

neutrinoceros (approved):
thank you !

PR #567, opened 2025-03-13
DEP: drop support for CPython 3.9

### LINKED ISSUE #519: DEP: Can we drop support for Python 3.9 ?
I'd like to propose dropping support for Python 3.9 now that we're about to add explicit support for 3.13.
My main motivation is that I'd like to have access to the `match/case` statement to implement support for `np.arange(...,  like=unyt_array(...))` which is part of #518. For reference, see my implementation for `astropy.units` in https://github.com/astropy/astropy/pull/17059

Thoughts, @jzuhone ?

neutrinoceros:
Close #519

### DIFF
--- a/pyproject.toml
+++ b/pyproject.toml
@@ -23,10 +23,10 @@ classifiers = [
 keywords = [
     "unyt",
 ]
-requires-python = ">=3.9"
+requires-python = ">=3.10"
 dependencies = [
-    "numpy>=1.19.3, <3.0",
-    "sympy>=1.7",
+    "numpy>=1.21.3, <3.0",
+    "sympy>=1.9.0",
     "packaging>=20.9",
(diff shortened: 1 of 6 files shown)


pyproject.toml
```diff
@@ -98,6 +98,8 @@ exclude = [
 ignore = [
     "E501",
     "B904",
+    "UP038", # non-pep604-isinstance
+    "B905", # zip-without-explicit-strict
```
neutrinoceros, line 102: this is temporary, I'll address this rule in a follow up PR

neutrinoceros: (#568)

neutrinoceros:
ping @jzuhone

jzuhone (approved):

PR #569, opened 2025-03-13
Add new authors for 3.0.4

jzuhone:


AUTHORS.rst
```diff
@@ -36,6 +36,8 @@ Contributors
 * Ben Thompson
 * Matthew Turk <matthewturk@gmail.com>
 * Miguel de Val-Borro <miguel.deval@gmail.com>
+* Jeroen Van Goey
+* Mike Zingale
```
neutrinoceros, line 40: If @zingale prefers Mike, we should also update HISTORY.rst, where I used Michael

zingale: no preference

PR #570, opened 2025-03-13
ENH: enable strict zip (almost) everywhere

### LINKED ISSUE #568: ENH: systematically use strict `zip`s
the builtin `zip` function gained a `strict` argument in Python 3.10, and ruff has a rule for checking that it's set (`B905`, or `zip-without-explicit-strict`), which even comes with autofix.
However, the auto-fix is just liberally sprinkling `strict=False` everywhere, when I believe the correct thing to do is almost always `strict=True`, so I'm choosing to ignore the rule in #567, but planning to fix it later on a case-by-case basis.

neutrinoceros:
based off #567
Close #568


unyt/dimensions.py
```diff
@@ -277,7 +277,9 @@ def new_f(*args, **kwargs):
                 If the units do not match.
 
             """
-            for arg_name, arg_value in chain(zip(names_of_args, args), kwargs.items()):
+            for arg_name, arg_value in chain(
+                zip(names_of_args, args, strict=False), kwargs.items()
```
neutrinoceros, line 281: this is the one place where I used `strict=False` because it makes sense in this context: when all arguments are passed as keywords, `args` is empty an d `names_of_args` is effectively the same as `list(kwargs.keys())`

PR #571, opened 2025-03-18
Don't use multiplication to attach units to result.

kyleaoman:
I realised that in [swiftsimio](https://github.com/SWIFTSIM/swiftsimio) we wanted to support `__mul__` and similar methods on our array class. This revealed that the implementation of `np.block` in unyt attaches units to the result by multiplying them on, which results in a partially initialised object calling `__mul__` on our side when we wrap `np.block`. This PR changes the wrapper for `np.block` to attach units using the `unyt_array` constructor, in line with other similar helper functions.


kyleaoman:
pre-commit.ci autofix

neutrinoceros (approved):
Thanks !

PR #572, opened 2025-03-19
Delegate unit multiplication when able

kyleaoman:
I have a subclass of `unyt_array` called `cosmo_array`. If I do `cosmo_array(...) * u.cm` the units multiply on as expected and I get back a result with `cosmo_array` type (and my subclass can do whatever it needs to in its `__mul__` method).

However, if I do `u.cm * cosmo_array(...)` then the units still multiply on as expected, but the result is silently demoted to a `unyt_array`. There is _nothing_ that can be done about this downstream because python checks the type of the left argument and first calls its `__mul__` method, only if that fails is control passed to the right argument's `__rmul__` method to see if the right argument knows how to handle the operation.

This PR proposes to let the right argument try to handle the unit multiplication if it knows how to by calling its `__rmul__` method. This resolved the issue with `u.cm * cosmo_array(...)` since my subclass implements a suitable `__rmul__`. It also does not seem to break any existing functionality.

This PR is a draft: I wanted to float the idea first before fleshing it out. The same treatment needs to be applied to `__truediv__`, for instance, and some test coverage is probably wanted.

Maybe breaking the conventional priority of `__mul__` over `__rmul__` in python is a horrific Thing That Should Not Be Done, I'm not sure, but would like to discuss.

### DIFF
--- a/unyt/unit_object.py
+++ b/unyt/unit_object.py
@@ -376,6 +376,12 @@ class Unit:
     def __mul__(self, u):
         """Multiply Unit with u (Unit object)."""
         if not getattr(u, "is_Unit", False):
+            try:
+                external_result = u.__rmul__(self)
+            except (TypeError, AttributeError):
+                pass
+            else:
+                return external_result
             data = np.array(u, subok=True)
             unit = getattr(u, "units", None)
             if unit is not None:


neutrinoceros:
This seems like a bad idea to me. I anticipate it could create a lot more confusion that it'd ever save, and I think I'd rather live with a non-symmetric mul operator.
There should be a better way. How about modifying the logic on lines 400-402 to allow for subclasses instead ?

kyleaoman:
@neutrinoceros that seems workable. Testing revealed that there's also a similar issue with some ufuncs, so I've also sketched in a possible handling for that... not sure yet it this makes sense, I thought that numpy would pass control to the sub-subclass even if it's the second argument of the ufunc, but it seems like no (specifically when mixing unyt_quantity and cosmo_array, which are both subclasses of unyt_array...).

Anyway, is this more or less an approach that can be considered? Would need refinement, proper testing, etc, just thinking conceptually at this stage.

kyleaoman:
pre-commit.ci autofix

kyleaoman:
@neutrinoceros I've tinkered a bit more and this is now enough to support what's wanted downstream. In brief it adds:
- for `unyt_subclass(...) * `u.cm` and `u.cm * `unyt_subclass` the subclass given the chance to handle the operation.
- for ufuncs, subclasses are given the option to implement two classmethods `__unyt_ufunc_prepare__` and `__unyt_ufunc_finalize__`. If they are present, they are called, passing all of the relevant information so that the subclass can prepare inputs and finalize outputs if it needs to.

I've implemented a use-case in swiftsimio (https://github.com/SWIFTSIM/swiftsimio/pull/221) and from that side everything works.

If this is an acceptable modification on the unyt side, then missing would be:
- consider whether any additional tests are wanted
- any further needed additions to documentation

Let me know what you think!

unyt/unit_object.py
```diff
@@ -371,12 +371,26 @@ def __itruediv__(self, u):
         )
 
     def __rmul__(self, u):
+        if not getattr(u, "is_Unit", False):
+            data = np.array(u, subok=True)
+            if (
+                issubclass(data.__class__, _import_cache_singleton.ua)
+                and type(data) is not _import_cache_singleton.ua
+                and type(data) is not _import_cache_singleton.uq
+            ):
```
neutrinoceros, line 380: Is there a reason to distinguish `data.__class__` and `type(data)` that I'm missing here ?
```suggestion
            data_type = type(data)
            if (
                issubclass(data_type, _import_cache_singleton.ua)
                and data_type is not _import_cache_singleton.ua
                and data_type is not _import_cache_singleton.uq
            ):
```

neutrinoceros (commented):
this seems like a good idea to me. Tests would indeed be needed, if nothing else, to ensure I understand how it's meant to be used.

kyleaoman:
Can do. I guess this means implementing a minimalist subclass and writing some tests around that?

neutrinoceros:
yep, that's how I'd think of it.

kyleaoman:
pre-commit.ci autofix

kyleaoman:
pre-commit.ci autofix

kyleaoman:
I adapted the swiftsimio `cosmo_array` and `cosmo_quantity` into a "minimal" pair of subclasses. They just have an extra attribute `extra_attr`, and the additional logic is pretty simple: the extra attribute can be `True`, `False` or `None` and if multiple operands have incompatible `extra_attr`, raise an exception `ExtraAttributeError`.

I also ported over and adapted all of the relevant tests of the subclasses from swiftsimio. You can for example comment out the `__unyt_ufunc_prepare__` and `__unyt_ufunc_finalize__` implementations and see that this causes a test failure. Same with the `__mul__` and `__rmul__` implementations. These are the new features for this PR, so I'd say that they're tested now.

There are a lot of tests and a lot of code in the subclass example - is this overkill? Not sure - to me it seems like if the claim is that supporting subclassing is desired, then having a full-fledged sample implementation that is reasonably well covered by its own test suite seems like a reasonable idea. See what you think...

kyleaoman:
The failing test is just a numpy function that didn't exist in older versions. I see that python3.9 support is slated to be dropped anyway - this will either just go away, or it's easy enough to code around it if you prefer.

jzuhone:
@kyleaoman I'll review this later today or tomorrow.

kyleaoman:
@jzuhone guess it didn't pan out last week, no problem. Just giving this a little nudge, if you can find some time at some point it's much appreciated.

jzuhone:
@kyleaoman looking at this now. we've gotten rid of python 3.9 support in `main`, can you rebase?

unyt/array.py
```diff
@@ -339,6 +339,67 @@ def _apply_power_mapping(ufunc, in_unit, in_size, in_shape, input_kwarg_dict):
...
+    by the subclass.
+
+    The __unyt_ufunc_finalize__ method should be a classmethod accepting
+    (result, ufunc, method, *inputs, **kwargs), similarly to above but with
+    the addition of result, the return value from unyt_array.__array_ufunc__.
+    It should return the result with any modifications needed by the subclass.
+    """
+
+    def wrapper(self, ufunc, method, *inputs, **kwargs):
+        if len(inputs) > 1:
```
jzuhone, line 366: Some comments in these lines as to what is being attempted here and what might raise `RuntimeError` would be helpful.

kyleaoman:
pre-commit.ci autofix

jzuhone:
@neutrinoceros can you have a look at the new tests here?

unyt/array.py
kyleaoman: Added some comments here.

kyleaoman:
> @kyleaoman looking at this now. we've gotten rid of python 3.9 support in `main`, can you rebase?

Done. Some issues around functions present/missing in different numpy versions coming up in the tests, will sort those out.

neutrinoceros:
wow, this blew up in size since the last time I had a look. 2k lines of code is a very different beast to review, and I certainly cannot commit to any schedule in this regard. Any chance this could be split into smaller bits somehow ?

kyleaoman:
@neutrinoceros 1000 lines is already just boilerplate implementing all of numpy. A lot of it is docstrings too. I could make the example more minimal if you think that's better, I just started with a pared down version of our existing subclass as the easiest option for me. Could you have an initial look at what's here (not the contents) and let me know what you think? If a full-fledged subclass is desired it's hard to make it much smaller. If something extremely minimal is better for some reason it could shrink to maybe a few hundred lines.

kyleaoman:
@neutrinoceros mmmk, I've cleaned up a few hundred lines that could unambiguously be dropped. Also cleaned up numpy version conflict stuff so the test suite passes.

From here a design decision is necessary. The example subclass implements the full range of numpy functions, and has some basic tests that this works. I could (almost?) completely remove this if you prefer. Some limited handling of numpy operations is probably still needed to support a minimally functional unyt_array subclass. Some limited support is also needed to demonstrate that the desired features of a fully-fledged subclass can be implemeted. At this point I'd ask for some guidance on how you think this should be approached conceptually.

kyleaoman:
@neutrinoceros would you be more in favour of a very minimalist implementation (getting rid of everything `@implements`'ed, for a start)?

kyleaoman:
@jzuhone just checking back - not trying to apply pressure, just keep it from falling off the radar.

jzuhone:
@kyleaoman I'm looking at this now.

jzuhone:
@kyleaoman did you happen to do any performance testing? Since we're doing extra checks in `__mul__` and `__rmul__`, as well as using the `_subclass_ufunc_prepare_and_finalize` function, I'd want to ensure that we don't take a big performance hit. 

I'm not saying we have any evidence that we would, but feel like we should check.

jzuhone:
@neutrinoceros what do you think about the timings here?

kyleaoman:
@neutrinoceros any thoughts on this in the end?

jzuhone:
@kyleaoman sorry for missing this--in the second plot, what do the curves refer to?

PR #574, opened 2025-04-03
DEP: declare test dependencies as PEP 735 dependency groups

neutrinoceros:
This is the fundamental piece from #532. I've stripped out the (partially broken) changes to CI, but this is already useful on it's own: it allows running the test suite with `uv run --group test pytest unyt` instead of a `tox` command.


neutrinoceros:
Actually de-duplication was easy. `tox-uv` and `tox-gh` are not needed for the basic stuff to work.

PR #584, opened 2025-07-26
MNT: adjust license spec to PEP 639

neutrinoceros:
`project.license` is deprecated as a TOML table since `setuptools` `77.0.0`:
https://setuptools.pypa.io/en/latest/history.html#id73
I adjusted for this change using the now prefered standard, [PEP 639](https://peps.python.org/pep-0639/#deprecate-license-key-table-subkeys), which also mentions the `License :: OSI Approved :: BSD License` classifier being ambiguous, and as such, leads to an error in `setuptools` if present, because it would override PEP 639 definitions, so I removed it.
Setuptools also auto-includes files listed in `project.license-files`, so we don't need to explicitly ask for it in `MANIFEST.in`

While I was at it, I noticed we were missing an obvious `Intended Audience` trove classifier, so I added it.

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,7 +1,6 @@
 include AUTHORS.rst
 include CONTRIBUTING.rst
 include HISTORY.rst
-include LICENSE
 include README.rst
 
 recursive-exclude * __pycache__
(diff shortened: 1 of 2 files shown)


neutrinoceros:
@jzuhone note that the associated deprecation warning from setuptools indicates that old-style license classifiers will break builds starting in next february, which makes this semi-urgent for the development version (users shouldn't really be affected as long as they consumed pre-built wheels)

PR #590, opened 2025-10-02
Update constants to CODATA 2022 Values

E-W-Jones:
After the yt workshop in Edinburgh I looked [this issue in yt](https://github.com/yt-project/yt/issues/5180) which stems from yt sometimes using its internal, and old, units, and sometimes unyt.

In doing so I discovered that a couple of unyt's constants could be updated to the CODATA 2022 values. I am getting confused by the conversion between SI and CGS units for electromagnetism, and so would appreciate someone else's eyes to tell me if Im missing something.

The issue stems from the fact that the SI had a shift in how they define/measure units, including the permeability of free space, $\mu_0$. It used to be _defined as exactly_ $4\pi \times 10^7~\mathrm{NA}^{-2}$. Now it has to be _measured_ as $1.25663706127(20)\times10^{-6}~\mathrm{NA}^{-2}$. This new value is within an error bar of the old one, and they differ at the 10th decimal place. This is useful because it means you _can_ get away with just not changing it.

When we update this value of $\mu_0$ it raises an error in testing our conversion between SI (metre, kilogram, second, coulomb) and CGS (centimetre, gram, second, statcoulomb) units. Specifically converting the B-field from tesla to gauss. To convert B field you introduce factors of $\sqrt{\frac{4\pi}{\mu_0}}$. This is how you get the equation of energy density used to test the conversion:
```
B = 1.0 * u.T
B_cgs = B.in_base("cgs")
u_mks = B * B / (2 * u.mu_0)
assert_equal(u_mks.units.dimensions, dimensions.pressure)
u_cgs = B_cgs * B_cgs / (8 * np.pi)
```

However, this doesn't happen when we convert from tesla to gauss. And it is only sometimes mentioned (for example not on the [Gauss (unit)](https://en.wikipedia.org/wiki/Gauss_(unit)) wikipedia page, but is on the [Gaussian units](https://en.wikipedia.org/wiki/Gaussian_units#Electromagnetic_unit_names) but in a way that is unclear to me if it part of the definition of the units. What I do know is that the ratio of `u_mks` to `u_cgs` when you convert them between unit systems is the same as the ratio of $\mu_0$ to $4\pi\times10^{-7}~\mathrm{NA}^{-2}$.


E-W-Jones:
An update in light of #636:

I have put the new values in and updated most tests to reflect this, currently there are two things left to do **which I need some input on**:
1. [test_old_registry_json](https://github.com/yt-project/unyt/blob/969fc6a629d332568832b849c6cb058d405ad442/unyt/tests/test_unit_registry.py#L165) looks like it justs reads in a saved registry and compares it to the default, but needed to check before updating the saved one with the new values.
        a. Is my interpretation correct, and
        b. is there a preferred way to do this, or is just editing the JSON fine?
2. CGS EM units are weird and the change to $\mu_0$ is small, I need someone to either
        a. say we should just ignore it (I have put a lot of work into not ignoring it so you know I'm not just being lazy suggesting it), or
        b. check my updated conversions are correct

To convert between CGS and SI units you are switching dimensions but you can work out the conversion between gauss and tesla (derived below, needs checking) as:
$$1\mathrm{G} \leftrightarrow \frac{10^4}{\sqrt{1+\delta}}\mathrm{T}$$
where $\delta = (-1.3\pm1.6)\times10^{-10}$ for the 2022 _measurement_ of $\mu_0$ (this was previously _defined_ with $\delta=0$).  This is consistent within 1 error bar with the previous definition.

I think it might be easier for everyone if we just ignore this. So a PR where we add all new values for every other constant, and ignore the 1e-10 error (that I believe will be subdominant in any cases of interest, as an example most tests are to 7 sig fig.). Perhaps I can add some notes to the [EM conversion section](https://github.com/yt-project/unyt/blob/main/unyt/unit_object.py#L800) about it.

Otherwise we will make changes like 1e4 -> 1.000000000066e4... that I really struggle to imagine makes any difference.

---
Conversion:
There are a couple of ways you could end up with
$$\frac{B_{CGS}}{B_{SI}} = \sqrt{\frac{4\pi}{\mu_0}}\quad(1)$$
but I like starting with magnetic energy density $\frac{B^2_{CGS}}{8\pi} = \frac{B^2_{SI}}{2\mu_0}$ and rearranging.

This constant on the RHS will (and you can check) have units $\mathrm{G}^{2} / \mathrm{T}^2$, and we can start by defining the new _measured_ $\mu_0 = 4\pi(1+\delta)\times 10^{-7}\mathrm{N}\mathrm{A}^{-2}$ where as a reminder $\delta\sim10^{-10}$.

So the LHS of (1) will evaluate to $$\sqrt{\frac{4\pi}{4\pi(1+\delta)\times 10^{-7}\mathrm{N}\mathrm{A}^{-2}}} = \sqrt{\frac{10^{7}\mathrm{A}^{2}\mathrm{N}^{-1}}{(1+\delta)}}\quad(2)$$

The next question is how to write amperes squared per newton in gauss per tesla:
$$\mathrm{G}^2 = \mathrm{erg}\mathrm{cm}^{-3} = 0.1\mathrm{N}\mathrm{m}^{-2}$$
$$\mathrm{T}^2 = \mathrm{N}^2\mathrm{A}^{-2}\mathrm{m}^{-2}$$
Divide these two by each other:
$$\frac{\mathrm{G}^2}{\mathrm{T}^2} = \frac{0.1\mathrm{N}\mathrm{m}^{-2}}{\mathrm{N}^2\mathrm{A}^{-2}\mathrm{m}^{-2}} = 0.1\mathrm{A}^{2}\mathrm{N}^{-1}$$
which is exactly what we want for (2)! Lets plug it in:
$$\frac{B_{CGS}}{B_{SI}} = \sqrt{\frac{10^{8}}{(1+\delta)}\frac{\mathrm{G}^2}{\mathrm{T}^2}} = \frac{10^{4}}{\sqrt{{1+\delta}}}\frac{\mathrm{G}}{\mathrm{T}}$$
which for the pre-2019 defintion of $\delta=0$ is what we would expect to convert between CGS and SI units.

chrishavlin:
just fyi @E-W-Jones , I pinged yt slack dev channel on your behalf, hoping someone with more domain expertise than myself can give you some feedback on your questions above!

jzuhone:
I will try to have a look at this during this week.

PR #594, opened 2025-10-08
TST: replace tox with uv as a command runner

### LINKED ISSUE #591: TST/CI: ditching tox in favor of uv ?
I would like to remove `tox` from our CI, and instead propose using `uv`, which I find solves actual problems with metadata duplication, whereas tox doesn't solve any actual issue I know of on its own. A possible middle ground is to use `tox-uv` (a tox extension that makes tox itself use uv), but I'd like to gather feedback first.

neutrinoceros:
close #591
replace #532

opening as a draft with minimal changes

TODO:
- [ ] make sure all tox environments are properly re-implemented as CI jobs
- [ ] udpate documentation (references to pyenv can be removed too)


CONTRIBUTING.rst
```diff
@@ -74,59 +74,36 @@ Get Started!
...
     $ pre-commit run --all-files
-    $ pytest --doctest-modules --doctest-rst --doctest-plus
-    $ pyenv local 3.10 3.11 3.12 3.13
-    $ tox
-    $ pyenv local 3.10
-
-   To get ``pre-commit``, ``pytest``, ``pytest-doctestplus``, and
-   ``tox``, just ``pip`` or ``conda`` install them into your python environment,
-   as appropriate. For a ``pyenv`` environment you would use ``pip``.
+    $ uv run --group test pytest --doctest-modules --doctest-rst --doctest-plus
```
chrishavlin: When I first ran this, I got 

```
======================================================== ERRORS ========================================================
__________________________________ ERROR collecting docs/extensions/show_all_units.py __________________________________
docs/extensions/show_all_units.py:1: in <module>
    from docutils.parsers.rst import Directive
E   ModuleNotFoundError: No module named 'docutils'
```
Should add the doc group I think? 

```suggestion
    $ uv run --group test --group doc pytest --doctest-modules --doctest-rst --doctest-plus
```

CONTRIBUTING.rst
```diff
@@ -179,8 +143,8 @@ Before you submit a pull request, check that it meets these guidelines:
 
 1. The pull request should include tests for functionality that is not already
    tested. We strive for 100% test coverage and pull requests should not add any
-   new untested code. You can generate coverage reports locally by running the
-   ``tox`` tests.
+   new untested code. You can generate coverage reports locally by running
+   ``pytest`` through ``coverage`` as ``uv run --group covcheck coverage run -m pytest``
```
chrishavlin: Tox used to create the html I think, the command as you have it will only create the `.coverage` statistics file, would be good to mention how to get human readable coverage stats: 

```suggestion
   new untested code. You can generate coverage reports locally by first generating coverage statistics with
   ``pytest`` through ``coverage`` as ``uv run --group covcheck coverage run -m pytest``. To generate reports from the coverage statistics, you can use ``uv run coverage report`` to print to screen or ``uv run coverage html`` to generate an html that you can open in a browser. 
```

PR #595, opened 2025-10-18
TST: fix incoming incompatibilities with numpy 2.4 (in1d is removed)

neutrinoceros:
xref https://github.com/numpy/numpy/pull/29978


neutrinoceros:
end users are not affected, but this might help downstream packagers, so I'd like to include this in our next patch release if possible @jzuhone

PR #597, opened 2025-10-19
TST: avoid mutating ndarray.shape attributes

neutrinoceros:
this is in anticipation of https://github.com/numpy/numpy/pull/29536


neutrinoceros:
Similar to #595 , this has no user-visible implication, but it could help downstream packagers, so I'd like to include this in the next patch release. @jzuhone

PR #610, opened 2025-12-22
Bugfix for histograms with mismatched units of input and bins

### LINKED ISSUE #609: `np.histogram` ignores units of bins
Here is a simple example of incorrect behaviour:
```
import numpy as np
import unyt as u

x = np.random.rand(100) * u.m
bins = np.linspace(0, 1, 6) * u.s
hist, resulting_bins = np.histogram(x, bins=bins)
assert np.all(bins == resulting_bins)
```
Here `resulting_bins` has units of `u.m`, and the `x` values are binned as though `bins = np.linspace(0, 1, 6) * u.m`. The incompatible units are silently ignored and replaced with whatever units `x` has.

kyleaoman:
Fixes #609

### DIFF
--- a/unyt/_array_functions.py
+++ b/unyt/_array_functions.py
@@ -826,8 +826,9 @@ def nanquantile(a, *args, **kwargs):
 
 @implements(np.linalg.det)
 def linalg_det(a, *args, **kwargs):
-    return np.linalg.det._implementation(np.asarray(a), *args, **kwargs) * a.units ** (
-        a.shape[0]
+    return (
+        np.linalg.det._implementation(np.asarray(a), *args, **kwargs)
+        * a.units ** (a.shape[0])
     )
 
 
--- a/unyt/tests/test_array_functions.py
+++ b/unyt/tests/test_array_functions.py
@@ -296,9 +296,9 @@ def test_wrapping_completeness():
     """Ensure we wrap all numpy functions that support __array_function__"""
     handled_numpy_functions = set(HANDLED_FUNCTIONS.keys())
     # ensure no functions appear in both NOT_HANDLED_FUNCTIONS and HANDLED_FUNCTIONS
-    assert NOT_HANDLED_FUNCTIONS.isdisjoint(
-        handled_numpy_functions
-    ), NOT_HANDLED_FUNCTIONS.intersection(handled_numpy_functions)
+    assert NOT_HANDLED_FUNCTIONS.isdisjoint(handled_numpy_functions), (
+        NOT_HANDLED_FUNCTIONS.intersection(handled_numpy_functions)
+    )
     # get list of functions that support wrapping by introspection on numpy module
     wrappable_functions = get_wrapped_functions(np, np.fft, np.linalg)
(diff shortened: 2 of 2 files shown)


neutrinoceros:
> any way to check if this patch at least maintains existing coverage without picking through line numbers "by hand"?

Yes ! for reference, you can download the full html coverage report from the Actions tab, and then visualize it in your local browser.

unyt/tests/test_subclass.py
```diff
@@ -727,7 +727,13 @@ def test_explicitly_handled_funcs(self):
             np.array([1, 2, 3]),
         ),
     )
-    @pytest.mark.parametrize("bins_type", ("int", "np", "sub_arr"))
+    @pytest.mark.parametrize(
+        "bins_type",
+        (
+            "int",
+            "sub_arr",
```
neutrinoceros:
```suggestion
            "sub_arr"
```

unyt/_array_functions.py
```diff
@@ -296,7 +329,12 @@ def _histogramdd(
 
     if weights is not None and hasattr(weights, "units"):
         counts *= weights.units
-    return counts, tuple(_bin * getattr(s, "units", 1) for _bin, s in zip(bins, sample))
+    if np.isscalar(bins):
+        return counts, tuple(b * getattr(s, "units", 1) for b, s in zip(_bins, sample))
```
neutrinoceros:
```suggestion
        return counts, tuple(b * getattr(s, "units", 1) for b, s in zip(_bins, sample, strict=True))
```

unyt/_array_functions.py
```diff
@@ -296,7 +329,12 @@ def _histogramdd(
...
+    return counts, tuple(
+        _b * getattr(s, "units", 1) if np.isscalar(b) else b
+        for b, _b, s in zip(bins, _bins, sample)
```
neutrinoceros:
```suggestion
        for b, _b, s in zip(bins, _bins, sample, strict=True)
```

unyt/_array_functions.py
```diff
@@ -158,20 +158,38 @@ def _sanitize_range(_range, units):
     return new_range.squeeze()
 
 
+def _sanitize_bins(a, bins):
+    if np.isscalar(bins):
+        # integer giving number of bins
+        return bins
```
neutrinoceros: it seems to me that a scalar with non-dimensionless units shouldn't be allowed. For sanity we should also strip the unit entirely if it's dimensionless.

unyt/_array_functions.py
```diff
@@ -206,20 +224,25 @@ def _histogram2d(x, y, *, bins=10, range=None, density=None, weights=None, norme
     range = _sanitize_range(
         range, units=[getattr(x, "units", None), getattr(y, "units", None)]
     )
+    if np.isscalar(bins):
+        _xbins, _ybins = _sanitize_bins(x, bins), _sanitize_bins(y, bins)
```
neutrinoceros: same here

PR #614, opened 2026-01-01
TST: adapt unyt_quantity.std test for forward compatibility with numpy 2.4 and newer

### LINKED ISSUE #603: BUG: incoming incompatibility with numpy-dev
[example logs](https://github.com/yt-project/unyt/actions/runs/18997393571/job/54259131034?pr=602)

The impacted test isn't released yet (source #593)
This is another blocker for our next patch release.

ping @jzuhone and @kyleaoman 

Keep in mind it's possible that the actual bug is upstream in numpy. I'll attempt a bisection to narrow it down.

neutrinoceros:
close #603
ping @kyleaoman for review

### DIFF
--- a/unyt/tests/test_unyt_array.py
+++ b/unyt/tests/test_unyt_array.py
@@ -3001,11 +3001,15 @@ def test_std_recursion():
     # see also https://github.com/yt-project/unyt/issues/540
     a = unyt_array([1, 1, 5, 5], 10 * m)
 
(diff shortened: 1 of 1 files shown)


neutrinoceros:
nah, xfails are treacherous (sometimes a test is broken for an unexpected reason but you'll never know until it's too late) so I prefer to avoid them if I can. Thank you for double checking !

PR #616, opened 2026-01-09
MNT: lock stable CI envs + schedule autoupgrades with dependabot

neutrinoceros:
The rationale is identical to https://github.com/yt-project/yt/issues/5310

PR #620, opened 2026-01-14
DOC: document how to bump version metadata with uv

neutrinoceros:
follow up to #619, after it was merged so fast I didn't get a chance to update it :)


jzuhone:
@neutrinoceros so if I have the version defined in those two files, `uv version bump` will fix them both?

neutrinoceros:
No. It'll only do pyproject.toml and uv.lock
But it's easy enough to implement a module level __version__ attribute that dynamically retrieves the metadata from the installed package if requested.

neutrinoceros:
(And that's exactly what I already did in #601)

neutrinoceros:
(and uv.lock is proposed for inclusion in #616)

PR #637, opened 2026-04-08
RFC: avoid deprecation warnings from numpy 2.5 when updating an array's dtype in place

### LINKED ISSUE #5420: TST: Upcoming dependency test failures
The weekly build with future dependencies has failed. Check the logs
https://github.com/yt-project/yt/actions/runs/24116908782

neutrinoceros:
close https://github.com/yt-project/yt/issues/5420

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -505,6 +505,25 @@ multiple_output_operators = {modf: 2, frexp: 2, divmod_: 2}
 LARGE_INPUT = {4: 16777217, 8: 9007199254740993}
 
 
+def _update_array_dtype_inplace(array, *, dtype) -> None:
+    # change the dtypes in-place, this does not change the
+    # underlying memory buffer
+    dt = np.dtype(dtype)
+    assert dt.kind == "f"
+    if NUMPY_VERSION >= Version("2.5.0dev0"):
+        array._set_dtype(dt)
+    else:
+        array.dtype = dt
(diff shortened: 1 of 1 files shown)


ngoldbaum:
Did you try deleting the code that updates the dtype and allow integer arrays through unchanged? Maybe it just works? 

Although it occurs to me now that maybe this code is setting up an in-place multiply that needs a float array, since multiplying by a non-integral conversion factor will suffer truncation error if the array stays an integer array.

Another possibly more disruptive option is to raise a TypeError so callers can reason about this.

neutrinoceros:
> Did you try deleting the code that updates the dtype and allow integer arrays through unchanged? Maybe it just works?

It doesn't. 15 tests break quite badly with errors like
```
numpy._core._exceptions._UFuncOutputCastingError: Cannot cast ufunc 'divide' output from dtype('float64') to dtype('int64') with casting rule 'same_kind'
```

> Another possibly more disruptive option is to raise a TypeError so callers can reason about this.

I agree that's probably where we should be going with this, though we could probably do that over a deprecation cycle.

neutrinoceros:
I'm thinking of just merging this as is to stabilize yt's bleeding edge CI before I get too accustomed to failures and stop paying attention to logs. I plan to do this next Tuesday if no objection is raised in the mean time.

PR #640, opened 2026-05-14
Allow other classes to handle multiplication & division with `Unit`

kyleaoman:
Currently if another class wants to implement multiplying with a `Unit`, it can only handle the case where it is the left argument, like:
```python
class MyClass:
    def __init__(self, x):
        self.x = x

    def __mul__(self, other):
        return self.x * other

    def __rmul__(self, other):
        return self.__mul__(other)

MyClass(5) * Unit("m")  # MyClass.__mul__ can handle this
```
When it is the right operand like this:
```python
Unit("m") * MyClass(5)
```
`Unit` raises a `InvalidUnitOperation` before `MyClass.__rmul__` ever has a chance to intervene.

When `Unit` can't resolve the operation itself, I'm proposing to ask the other argument whether it knows what to do and returning the resulting value if it does, otherwise `Unit` will raise as before. This is labelled "Option A" in the proposed changes and keeps the new behaviour as close as possible to the current behaviour at the cost of a more complex implementation.

Alternatively there is "Option B", currently commented out, that would remove the `raise` and simply `return NotImplemented` when `Unit` doesn't know what to do. This is more in line with how a python class is expected to behave but changes how `Unit` behaves.

The included tests have a bare-bones example of how a class might use this functionality.

#641  is related.

### DIFF
--- a/unyt/unit_object.py
+++ b/unyt/unit_object.py
@@ -402,10 +402,21 @@ class Unit:
             else:
                 units = self
             if data.dtype.kind not in ("f", "u", "i", "c"):
+                # ---------------- OPTION A ---------------
+                if hasattr(u, "__mul__"):
+                    try:
+                        ret = u.__mul__(self)
+                        if ret is not NotImplemented:
+                            return ret
+                    except TypeError:
+                        pass
                 raise InvalidUnitOperation(
                     f"Tried to multiply a Unit object with '{u}' (type {type(u)}). "
                     "This behavior is undefined."
                 )
+                # ---------------- OPTION B ---------------
+                # return NotImplemented
+                # -----------------------------------------
             if data.shape == ():
                 return _import_cache_singleton.uq(data, units, bypass_validation=True)
             return _import_cache_singleton.ua(data, units, bypass_validation=True)
@@ -442,10 +453,21 @@ class Unit:
 
                 return unyt_quantity(1.0, self) / u
             else:
+                # ---------------- OPTION A ---------------
+                if hasattr(u, "__rtruediv__"):
+                    try:
+                        ret = u.__rtruediv__(self)
+                        if ret is not NotImplemented:
+                            return ret
+                    except TypeError:
+                        pass
                 raise InvalidUnitOperation(
                     f"Tried to divide a Unit object by '{u}' (type {type(u)}). "
                     "This behavior is undefined."
                 )
+                # ---------------- OPTION B ---------------
+                # return NotImplemented
+                # -----------------------------------------
         elif self.dimensions is logarithmic and not u.is_dimensionless:
             raise InvalidUnitOperation(f"Tried to divide '{self}' and '{u}'.")
         elif u.dimensions is logarithmic and not self.is_dimensionless:
--- a/unyt/unit_object.py
+++ b/unyt/unit_object.py
@@ -403,19 +403,19 @@ class Unit:
                 units = self
             if data.dtype.kind not in ("f", "u", "i", "c"):
                 # ---------------- OPTION A ---------------
-                if hasattr(u, "__mul__"):
-                    try:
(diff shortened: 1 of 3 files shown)


kyleaoman:
@jzuhone @neutrinoceros could I give this (& #641) a little bump?

unyt/unit_object.py
```diff
@@ -402,10 +402,21 @@ def __mul__(self, u):
...
+                        ret = u.__mul__(self)
+                        if ret is not NotImplemented:
+                            return ret
+                    except TypeError:
+                        pass
                 raise InvalidUnitOperation(
                     f"Tried to multiply a Unit object with '{u}' (type {type(u)}). "
                     "This behavior is undefined."
                 )
+                # ---------------- OPTION B ---------------
```
jzuhone: @kyleaoman do we need this commented-out part in, as it doesn't do anything?

jzuhone: Sorry, didn't read the description closely enough. My bad!

jzuhone:
@kyleaoman I have a question about this, similar to a discussion we had on a previous PR. Did you happen to do a timing study on these changes? As I've mentioned before, we want to make sure when we change basic unit or array/quantity operations that we don't impose a speed penalty. 

It looks from the design of your changes that this doesn't apply in the vast majority of cases, but it would be good to make this a standard test.

Also, I guess I don't understand what the point of Option B is--as it's written, you would never get to this code because you would raise the `InvalidUnitOperation` first without getting to Option B.

kyleaoman:
Hi @jzuhone,

I'm happy to investigate timing after some initial discussion of the two options. To clarify, Option B replaces all of the code between the Option A comment and the Option B comment (so L406 to L416 are removed in this scenario). You just go straight to `return NotImplemented`.

Currently you're using `raise InvalidUnitOperation` analogously to `raise NotImplementedError`. If you check the docs for `NotImplemented` and `NotImplementedError`, this approach is incorrect because it prevents the other argument from resolving the operation:
https://docs.python.org/3/library/constants.html#NotImplemented
https://docs.python.org/3/library/exceptions.html#NotImplementedError
So I think that Option B is very strongly preferable, the only drawback is that it's an API change (but arguably one that doesn't matter much, if the other argument fails to handle the operation you still raise, just `TypeError` instead of `InvalidUnitOperation`).

kyleaoman:
To make it concrete I've swapped the implementation over to "option B", which requires a few adjustments to the tests to expect `TypeError` instead of `InvalidUnitOperation` in some places. Can always roll back to "option A" if that's preferred.

jzuhone:
@kyleaoman I definitely agree that Option B is to be preferred here, so let's go with it.

kyleaoman:
Ok! Does this really need performance testing then? Really just substituting return NotImplemented for raise statements in two places... Happy to do it if you think it's needed though.

jzuhone:
If we go with Option B, we don't need the performance testing.

kyleaoman:
@jzuhone can this be merged, then?

unyt/unit_object.py
```diff
@@ -442,10 +439,21 @@ def __truediv__(self, u):
 
                 return unyt_quantity(1.0, self) / u
             else:
-                raise InvalidUnitOperation(
-                    f"Tried to divide a Unit object by '{u}' (type {type(u)}). "
-                    "This behavior is undefined."
-                )
+                # ---------------- OPTION A ---------------
```
jzuhone: @kyleaoman can we add a descriptive comment here and get rid of the commented-out code?

jzuhone (changes requested):

unyt/unit_object.py
kyleaoman: Whoops had forgotten to tidy up this second block. Done now.

PR #641, opened 2026-05-14
Allow other (non-sub)classes to handle binary ufuncs with `unyt_array`

kyleaoman:
#572 added the ability for subclasses of `unyt_array` to intervene before and after numpy ufunc calls to "prepare" the arguments and "finalize" the result.

In this PR I propose to extend this ability to classes that are not subclasses, allowing them to implement `__unyt_ufunc_prepare__` and `__unyt_ufunc_finalize__` if they wish. Note that the other class cannot control the result via e.g. `__mul__` and `__rmul__` because control ends up in `unyt_array.__array_ufunc__` and leads to an exception leaving the other class no chance to intervene.

The included tests have a bare-bones example of how a class might use this functionality.

#640 is related.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -339,7 +339,7 @@ def _apply_power_mapping(ufunc, in_unit, in_size, in_shape, input_kwarg_dict):
     return mul, unit
 
 
-def _subclass_ufunc_prepare_and_finalize(ufunc_handler):
+def _ufunc_prepare_and_finalize(ufunc_handler):
     """
     This wrapper function is intended to be applied to __array_ufunc__ on
     unyt_array.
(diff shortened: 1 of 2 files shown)


jzuhone:
I guess I would like a more detailed explanation of why this might be a better way to handle things instead of just making the "other" class you want to operate with a subclass of `unyt_array` or `unyt_quantity`. 

I mentioned in https://github.com/yt-project/unyt/pull/640 that we should check if this has any negative impacts on performance, I'd like that check to be done here as well.

kyleaoman:
Some further research led me to a much simpler proposal, picking this up in #674

PR #662, opened 2026-09-01
API: drop support for most operations on integral dtype unyt arrays

### LINKED ISSUE #598: API: what should we do about in-place mutator methods that change an array's dtype ?
### Description

upstream context: https://github.com/numpy/numpy/issues/28800

in short: modifying an array's `dtype` attribute in place is planned for a deprecation in some future version of numpy.
We do have a couple public APIs where this is used:
1) `unyt_array.convert_to_*` functions, which are explicitly documented as doing in-place modifications)
2) `unyt_array.__array_ufunc__` has special logic for handling an `out` argument with an integer dtype

In the latter, I don't think we have any options other than dropping support for this special, ill defined case.
The former however could go different ways:
- pure and simple deprecation, with a warning saying that this *will* break for sufficiently recent versions of numpy (seems impossible given the amount of code that relies on these methods, in yt alone)
- renounce in-place mutation and instead produce copies, unconditionally (seems like violating a promise)
- add a new argument for handling cases where in-place conversions cannot be done without mutating `dtype`. Options could be:
  - do everything in place whenever possible, but allow copies if necessary
  - always return a copy
  - never return a copy, error if this constraints cannot be satisfied

Note that the last option (a new keyword argument) matches how the `copy` keyword argument works in many APIs in numpy 2.0+ (options are `None` (default), `True` and `False`), so it should be pretty straight forward to understand for new users.
What to you think @jzuhone ?

neutrinoceros:
Rationale: in-place mutations to unerlying dtypes are a can of worm, and most of the time our special-casing for integral dtypes is surprising and insufficient, so I propose we simply stop trying and focus our efforts on what actually matters: floating point arithmetics.

This is the aggressive way to close #598, and obviously consistute a breaking change so, if approved, this would mandate a major version bump.


neutrinoceros:
ping @jzuhone for review
this was originally suggested by @ngoldbaum and I took it to the very extreme.

jzuhone:
@neutrinoceros in principle I am for this. I need to look at why we did all this in the first place.

jzuhone:
See https://github.com/yt-project/unyt/pull/56 for context.

jzuhone:
@neutrinoceros I must say, however, that the convenience of writing something like `ds.sphere("c", (200, "kpc"))` without having to remember to make it `200.0` all of the time is probably pretty widespread among users (I can't tell you how much of my code would break if we didn't allow this). I'm sorry I haven't reviewed yet in detail, but I would be in favor of preserving the ability to create new arrays or quantities using lists of ints and single ints (respectively), but dropping support for in-place conversions of existing arrays. Despite the recent emphasis on type hinting and the like, the vast majority of users simply do not see something like `200` as an `int` but instead as just a number they want to use in a calculation.

neutrinoceros:
I actually agree, the amount of damage is probably not acceptable. One alternative would be to just silently convert integral inputs to floats, though there's probably another subset of our users that this could hurt...

jzuhone:
Or maybe not?

What if we handled it such that for lists of ints or single ints they got converted to float64, but ndarrays keep their dtype and that dtype cannot be changed in-place?

PR #674, opened 2026-09-02
Allow other (non-sub)classes to handle binary ufuncs with `unyt_array`

kyleaoman:
Following some further research and tinkering, this is a second attempt at #641 that is a *much* simpler implementation. It removes a lot of logic that subclasses previously had to hook into, instead `unyt_array.__array_ufunc__` now returns `NotImplemented` when it is confronted with (1) input that it cannot understand, or (2) a subclass (except `unyt_quantity`). This allows both subclasses and other classes to have a chance to decide how the operation should be resolved before unyt unilaterally raises an exception. The previous tests still pass, and I've added a new one illustrating how a class that is not a subclass can make use of `NotImplemented` being returned to implement right-sided multiplication and division.

### DIFF
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -1794,13 +1794,13 @@ class unyt_array(np.ndarray):
     def __eq__(self, other):
         try:
             return super().__eq__(other)
-        except (IterableUnitCoercionError, UnitOperationError):
+        except (IterableUnitCoercionError, UnitOperationError, TypeError):
             return np.zeros(self.shape, dtype="bool")
 
     def __ne__(self, other):
         try:
             return super().__ne__(other)
-        except (IterableUnitCoercionError, UnitOperationError):
+        except (IterableUnitCoercionError, UnitOperationError, TypeError):
             return np.ones(self.shape, dtype="bool")
 
     #
--- a/unyt/array.py
+++ b/unyt/array.py
@@ -339,73 +339,6 @@ def _apply_power_mapping(ufunc, in_unit, in_size, in_shape, input_kwarg_dict):
     return mul, unit
 
 
-def _subclass_ufunc_prepare_and_finalize(ufunc_handler):
-    """
-    This wrapper function is intended to be applied to __array_ufunc__ on
-    unyt_array.
-
-    It enables subclasses of unyt_array and unyt_quantity to define two
-    classmethods: __unyt_ufunc_prepare__ and __unyt_ufunc_finalize__.
-    These are called if available to allow subclasses to prepare arguments
-    for and modify return values from numpy ufuncs.
-
-    The __unyt_ufunc_prepare__ method should be a classmethod accepting
-    (ufunc, method, *inputs, **kwargs), i.e. the same inputs as
-    __array_ufunc__. It should return
-    (prepared_ufunc, prepared_method, prepared_inputs, prepared_kwargs)
-    which correspond logically to its inputs with any modifications needed
-    by the subclass.
-
-    The __unyt_ufunc_finalize__ method should be a classmethod accepting
-    (result, ufunc, method, *inputs, **kwargs), similarly to above but with
-    the addition of result, the return value from unyt_array.__array_ufunc__.
-    It should return the result with any modifications needed by the subclass.
-    """
-
-    def wrapper(self, ufunc, method, *inputs, **kwargs):
-        if len(inputs) > 1:
-            try:
-                # check if we're dealing with types that we understand and get
-                # the return type, otherwise _get_binary_op_return_class will
-                # raise RuntimeError.
-                ret_class = _get_binary_op_return_class(
-                    type(inputs[0]), type(inputs[1])
-                )
-            except RuntimeError:
-                # we're sure none of the arguments want to modify the input
-                # or output, so bypass __unyt_ufunc_prepare__ and
-                # __unyt_ufunc_finalize__.
-                return ufunc_handler(self, ufunc, method, *inputs, **kwargs)
-        else:
-            ret_class = type(inputs[0])
-        if hasattr(ret_class, "__unyt_ufunc_prepare__"):
-            prepared_ufunc, prepared_method, prepared_inputs, prepared_kwargs = (
-                ret_class.__unyt_ufunc_prepare__(ufunc, method, *inputs, **kwargs)
-            )
-        else:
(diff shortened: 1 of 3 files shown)


kyleaoman:
To give a bit more explanation as to why this is desirable, let's say I have a new class that wants to interact with `unyt_arrays`:
```python
class MyClass:
        def __init__(self, x: int):
            self.x = x

        def __mul__(self, b):
            return self.x * b

        def __rmul__(self, b):
            return self.x * b

        def __truediv__(self, b):
            return self.x / b

        def __rtruediv__(self, b):
            return b / self.x

        __array_ufunc__ = None  # this is here as a signal to the numpy dispatcher
```
The class just stores an integer and declares how multiplication and division with the class should proceed: you just get the result of the integer multiplied or divided as appropriate. So `MyClass(5) * 3` evaluates to `15`, simple enough. However, it also defines the operations when it is the right-hand argument, so that even though a basic integer doesn't know how to multiply with `MyClass`, we still have `3 * MyClass(5)` evaluate to `15`. This is because `int(3).__mul__(MyClass(5))` returns `NotImplemented`, so python then goes to the right hand argument and tries `MyClass(5).__rmul__(3)`, which evaluates to `15`.

If we try the same with a `unyt_array`, let's say `ua = [3] * u.m`, then when `MyClass` is the left argument we go straight to `MyClass.__mul__` and everything works: `MyClass(5) * ua` evaluates to `unyt_array([15], 'm')`. However if we flip the argument order to `ua * MyClass(5)` we go to `unyt_array.__array_ufunc__` first. That method can't figure out what to do with `MyClass` so it *raises* `unyt.exceptions.IterableUnitCoercionError`. With that raise, `MyClass` is completely locked out of intervening to declare that it knows what to do. The solution is to `return NotImplemented` instead of raising. This signals to the numpy and python dispatchers that `unyt_array` doesn't know how to resolve the operation but lets the other argument intervene, if it can.

That's the gist of what this PR implements. With those changes, the above example `ua * MyClass(5)` evaluates to `unyt_array([15], 'm')`, as desired. As a bonus, we can tidy up some bloated code (`_subclass_ufunc_prepare_and_finalize`) that isn't needed if `unyt_array` conforms more to how classes are "supposed to" behave (https://docs.python.org/3/library/constants.html#NotImplemented).

I can do some performance tests, but would be appreciated if we could discuss the general conceptual approach first - if the implementation details change even slightly performance tests become void...

kyleaoman:
There are a couple of test failures, seems that I've broken string comparison which probably points to a little bit more care being needed when deciding to `return NotImplemented`. Will have a look when I get a chance, need to head off now.

kyleaoman:
Those test failures are sorted, so worth an initial look at this point, I'd suggest. It's fairly straightforward in this new implementation.

kyleaoman:
@jzuhone Just giving this a little bump. This underpins a feature we've been hoping to add for a while in swiftsimio. We have a collaboration meeting in about a month, would be great if it was possible for us to annouce that new feature there, provided you and/or @neutrinoceros have time to review/release on that time scale.

jzuhone:
@kyleaoman since this one isn't as straightforward, can we do the performance test here?

kyleaoman:
I had a go at performance testing and my conclusion is that there's no measurable performance impact.

If you look at the functional changes to `__array_ufunc__`, they are quite minimal, so this makes sense.

I tried things like this:
```python
import timeit
import numpy as np
import unyt as u

N = 1000000
np.random.seed(0)
arr1 = np.random.rand(N)
arr2 = np.random.rand(N)
uarr1 = arr1 * u.m
uarr2 = arr2 * u.s

tarr = timeit.timeit("arr1 * arr2", number=10000, globals=globals())
tuarr = timeit.timeit("uarr1 * uarr2", number=10000, globals=globals())

print(f"numpy takes {tarr:.5f}s")
print(f" unyt takes {tuarr:.5f}s")
print(
    f"unyt {100 * np.abs(tarr - tuarr) / tarr:.2f}% "
    f"{'faster' if tarr > tuarr else 'slower'} than numpy"
)
```
With large `N` (say 1 million) the calculation is dominated by the actual array multiplication. I've run it on both branches and can't tell them apart. With both branches `unyt` even sometimes "beats" `numpy` (by which I mean the difference is less than the noise in the measurement).

With small `N` (say 10) the `unyt` overheads become important and `unyt` is consistently slower than `numpy`. However which branch is faster is random run-to-run, again the difference is smaller than the noise in the measurement.

unyt/array.py
```diff
@@ -1861,21 +1794,27 @@ def __pow__(self, p, mod=None, /):
...
+        except (IterableUnitCoercionError, UnitOperationError, TypeError):
             return np.ones(self.shape, dtype="bool")
 
     #
     # Start operation methods
     #
 
-    @_subclass_ufunc_prepare_and_finalize
     def __array_ufunc__(self, ufunc, method, *inputs, **kwargs):
+        if type(self) in (unyt_array, unyt_quantity):
```
jzuhone, line 1811: @kyleaoman sorry, one last thing--just leave a comment for what this is checking.

jzuhone (changes requested):
Hi @kyleaoman I left one thing to change, and then I think I can merge.

Lastly, what's more important then demonstrating the speed against numpy is demonstrating that the performance is not impacted between the tip of `main` and your proposed changes. Maybe you looked at that, but it isn't what you mentioned in your comments.

kyleaoman:
Added a comment as suggested.

Re the speed testing, sorry if my commentary wasn't clear. I was finding that which of the two branches was faster was seemingly random from run to run. I added the pure numpy case to give me a clear expectation: unyt should never be faster than the equivalent pure numpy operation. With that, I was finding that both branches could sometimes be faster than pure numpy. Since I don't believe that reflects a real performance edge, my conclusion is that the performance is basically identical up to my ability to measure it due to "noise" in the measurement (even averaging over 10s of thousands of iterations).

kyleaoman:
If it were possible to release an update in the not-too-distant future once this is merged (and probably best to wait for the weekly (?) yt tests to run), that would be great.

neutrinoceros:
> to wait for the weekly (?) yt tests to run


just flying by: yes, it's weekly (wednesday), but I'll run it manually now

neutrinoceros:
see: https://github.com/yt-project/yt/actions/runs/35616949866

Issue #577, opened 2025-04-23
unexpected behavior: convert delta_degC to degC and vice versa


Cpp87:
Converting delta_degC to degC implicitly assumes the base value 0 Kelvin. Example:

```py
>>> (3*unyt.delta_degC).to(unyt.degC)
unyt_quantity(-270.15, 'degC')
```

The offset of 273.15 is suprising to me. I would rather expect an error.

How did I encounter this:
I did not know about the delta_degC. My code calculates in Celcius and I perform x.to(unyt.degC) in case the user inputs Kelvin.
This would have produced terribly wrong results. Hence, I suggest to raise an error and tell the user "provide base temperture for converting delta_degC to degC". Implicit assumption of 0K as base temperature is dangerous.

The same issue happens vice versa:
```py
>>> (3*unyt.degC).to(unyt.delta_degC)
unyt_quantity(276.15, 'delta_degC')
```

Side note:
Substracting two temperatures in Celcius yields delta_degC as unit.
Substracting two temperatures in Kelving yields Kelvin - seems inconsistent.

Cpp87:
Unyt version: 3.0.3
Phyton version: 3.10.16
OS: Windows

neutrinoceros:
> I'm however not sure about your proposed solution

to be more actionable: I do not mean that this is *definitely* not feasible, but I couldn't convince myself it was feasible by studying the code base for 10min. I would be happy to be persuaded otherwise. A proof-of-concept level PR might do the trick.

PR #676, opened 2026-09-18
Refuse converting interval temperature to offset temperature

shaneraphel:
Related to #577.

`(3 * delta_degC).to(degC)` applies the 0 K affine offset and returns `-270.15 °C`. An interval and a point on the affine Celsius line are not the same quantity.

Public conversion (`in_units` / `convert_to_units`, which `.to` calls) now raises when an offset is about to be applied and either side is a `delta_` unit. Addition is unchanged: `10*degC + 1*delta_degC` is still `11*degC`, because that path reads only the scale from `get_conversion_factor`. Kelvin ↔ Celsius and Celsius ↔ Fahrenheit stay.

`unyt/tests/test_temperature_interval_offset.py` covers both directions for °C and °F, Kelvin ↔ Celsius, and interval → Kelvin.

AI disclosure: I used AI assistance while drafting this change, and I have reviewed all of the code in it.


unyt/tests/test_temperature_interval_offset.py
```diff
@@ -0,0 +1,35 @@
...
+def test_kelvin_celsius_affine_stays():
+    assert (0 * degC).to(K) == 273.15 * K
+    assert (273.15 * K).to(degC) == 0 * degC
+
+
+def test_interval_to_kelvin_stays():
+    assert (3 * delta_degC).to(K) == 3 * K
+
+
+def test_leftover_offset_identity():
```
jzuhone: This is not a useful test, please remove it. All it does is check that Python subtraction works.

unyt/tests/test_temperature_interval_offset.py
```diff
@@ -0,0 +1,35 @@
...
+        (3 * delta_degC).to(degC)
+    with pytest.raises(InvalidUnitOperation):
+        (3 * degC).to(delta_degC)
+    with pytest.raises(InvalidUnitOperation):
+        (3 * delta_degF).to(degF)
+    with pytest.raises(InvalidUnitOperation):
+        (3 * degF).to(delta_degF)
+
+
+def test_affine_plus_interval_stays():
```
jzuhone: This test is not needed--because we are already testing it. Please see:

https://github.com/yt-project/unyt/blob/a9e68a3926d0e330498fd8ed1ca87836600eaf65/unyt/tests/test_unyt_array.py#L2529-L2539

https://github.com/yt-project/unyt/blob/a9e68a3926d0e330498fd8ed1ca87836600eaf65/unyt/tests/test_unyt_array.py#L2542-L2552

jzuhone (changes requested):
@shaneraphel I'm in principle in favor of this PR to fix the issue, but I noticed that one of the proposed tests is redundant and another one is only testing python subtraction--not sure why this is even necessary.

Please also acknowledge the use of AI in the pull request, and if AI was used that you reviewed all of the code that was produced.

shaneraphel:
thanks for review,i will fix it.

unyt/tests/test_temperature_interval_offset.py
shaneraphel: Removed.

unyt/tests/test_temperature_interval_offset.py
shaneraphel: Removed — you're right, affine-plus-interval addition is already covered in test_unyt_array.py.

shaneraphel:
Thanks for the review. Done in the latest push.

PR #677, opened 2026-09-18
BUG: properly return NotImplemented in array interoperability methods

ikrommyd:
Makes the array interoperability functions like `__array_ufunc__` and `__array_function__` properly `return NotImplemented` when they don't support a ufunc/function.

Discovered while doing testing with external ndarray subclasses for https://github.com/numpy/numpy/pull/32577

Without this, unyt raised a KeyError
```
File ~/repos/unyt/unyt/array.py:1914, in unyt_array.__array_ufunc__(self, ufunc, method, *inputs, **kwargs)
   1911     mul, unit = _apply_power_mapping(ufunc, u, inp.size, inp.shape, kwargs)
   1912 else:
   1913     # get unit of result
-> 1914     mul, unit = self._ufunc_registry[ufunc](u)
   1915 # use type(self) here so we can support user-defined
   1916 # subclasses of unyt_array
   1917 ret_class = type(self)

KeyError: <ufunc 'minimummaximum'>
```
while the spec says you should reuturn NotImplemented and then you get this proper TypeError from numpy
```
TypeError: operand type(s) all returned NotImplemented from __array_ufunc__(<ufunc 'minimummaximum'>, 'reduce', unyt_array([1, 2, 3], 'cm'), axis=None, dtype=None): 'unyt_array'
```


ikrommyd:
Pinging @neutrinoceros, this is not urgent but it might bite anyone using numpy from main (after we merge the PR linked in the description) or when numpy 2.6.0 comes out cause we will expect that libraries that implement ndarray subclasses follow the spec regarding these methods.

